PR #3, opened 2022-08-08
WIP: backend: Fix ax handshaking req_ready does no longer depend on ax_ready

thommythomaso:


VERSION
```diff
@@ -1 +1 @@
-0.2.1
+0.3.0-dev
```
micprog, line 1: As this is a backwards-compatible fix, I believe a 0.2.2 version should be sufficient.

micprog (commented):
I am not 100% sure, but there may be some further incorrect handshake dependencies in idma_legalizer: https://github.com/pulp-platform/iDMA/blob/fix/backend/ax_handshaking-tbenz/src/idma_legalizer.sv#L379-L386
Another may be in idma_axi_transport layer: https://github.com/pulp-platform/iDMA/blob/fix/backend/ax_handshaking-tbenz/src/idma_axi_transport_layer.sv#L249-L257
I might be wrong, and I don't believe these affect the AXI path, but would be great if you could look at it quickly as well.

micprog:
> I agree to lines #L379-L386 in the legalizer. This has to be rewritten at some point (this is anyway planned soon). So far this has never caused any issues so let's create an issue and postpone rewriting until a bug manifests.
> The datapath, although looking a bit weird, should be correct.

Ok, thanks for checking!

micprog (approved):
Approve PR pending creation of new issue on ready-valid dependency in legalizer

thommythomaso:
Issue [#6](https://github.com/pulp-platform/iDMA/issues/6) created.

PR #5, opened 2022-08-08
Initial github CI

micprog:
Move open-source CI components to github actions


Makefile
```diff
@@ -54,7 +54,7 @@ XVLOG_ARGS += -64bit -compile -vtimescale 1ns/1ns -quiet
 
 define generate_vsim
 	echo 'set ROOT [file normalize [file dirname [info script]]/$3]' > $1
-	bender script vsim --vlog-arg="$(VLOG_ARGS)" $2 | grep -v "set ROOT" >> $1
+	$(BENDER) script vsim --vlog-arg="$(VLOG_ARGS)" $2 | grep -v "set ROOT" >> $1
```
thommythomaso, line 57: Is `bender` here as a default correct? or should the default be `./bender`?

Makefile
```diff
@@ -258,6 +258,17 @@ gen_ci: .gitlab-ci.yml
 ci_clean:
 	rm -f .gitlab-ci.yml
 
+bender:
```
thommythomaso, line 261: Is this part of the `CI`-section of the makefile? or should we put this in a dedicated section?

thommythomaso (commented):
Awesome work! Thanks @micprog for allowing the checks and the doc to run in the GitHub CI.

Makefile
micprog: Currently this is only used in the CI, but I'm open to moving it to the main section and updating the `BENDER` variable on top. Updating the top variable would probably also be improved if the necessary make targets have the bender target as a prerequisite.

thommythomaso (approved):

PR #8, opened 2022-08-11
testbench: Add capabilities to add non-idealities to the simulation memory

thommythomaso:


micprog (approved):
LGTM, although I suggest to externalize the axi throttling into an `axi_throttle` module (or similar) to avoid code duplication.

thommythomaso:
> LGTM, although I suggest to externalize the axi throttling into an `axi_throttle` module (or similar) to avoid code duplication.

Done. I created the `axi_throttle`module. Can you have a quick peek at it?

test/axi_throttle.sv
```diff
@@ -0,0 +1,101 @@
...
+/// indistinguishability of the request/responses (all ARs and AWs have the same ID respectively).
+module axi_throttle #(
+    /// The maximum amount of allowable outstanding write requests
+    parameter int unsigned MaxNumAwPending = 1,
+    /// The maximum amount of allowable outstanding read requests
+    parameter int unsigned MaxNumArPending = 1,
+    /// AXI4+ATOP request type
+    parameter type req_t = logic,
+    /// AXI4+ATOP response type
+    parameter type rsp_t = logic,
```
micprog, line 19: I suggest to use `axi_req_t` and `axi_rsp_t`.

From https://github.com/pulp-platform/axi/issues/153 :
> If the name of a type parameter does not have a port-specific prefix, it MUST be prefixed with axi_.
>Rationale: Some tools do not properly scope type definitions, and adding a topic-specific prefix reduces the chance of type name collisions.

micprog (approved):
LGTM other than the minor thing noted, thanks for the quick fix!

test/axi_throttle.sv
thommythomaso: Good catch! I will do this an merge it.

PR #15, opened 2022-10-27
dma_core_wrap, idma_reg64_frontend: Minor fixes

Vyeow:
In dma_core_wrap we encountered this error:
`Error: src/systems/cva6_reg/dma_core_wrap.sv(60): 'burst_req_t' is an unknown type.`
We believe that this was meant to be idma_req_t as this type was declared just before.

And in idma_reg64_frontend we encountered these warnings:
```
Warning: (vsim-3015) [PCDPC] - Port size (64) does not match connection size (48) for port 'next_o'. The port definition is at: src/frontends/idma_transfer_id_gen.sv(19).
File: src/frontends/register_64bit/idma_reg64_frontend.sv Line: 143

Warning: (vsim-3015) [PCDPC] - Port size (64) does not match connection size (48) for port 'completed_o'. The port definition is at: src/frontends/idma_transfer_id_gen.sv(21).
File: src/frontends/register_64bit/idma_reg64_frontend.sv Line: 143
```
The signals next_id, done_id are declared with length DmaAddrWidth instead of DmaRegisterWidth. We believe that this is unintentional, as the config registers are 64 bits wide.
However, note that with this fix the parameter DmaAddrWidth will not be used anywhere in the module.

### DIFF
--- a/src/frontends/register_64bit/idma_reg64_frontend.sv
+++ b/src/frontends/register_64bit/idma_reg64_frontend.sv
@@ -7,8 +7,6 @@
 // Description: DMA frontend module that includes 64bit config and status reg handling
 
 module idma_reg64_frontend #(
-    /// address width of the DMA AXI Master port
-    parameter int  unsigned DmaAddrWidth     = -1,
     /// register_interface request type
     parameter type          dma_regs_req_t   = logic,
     /// register_interface response type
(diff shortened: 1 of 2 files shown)


thommythomaso (approved):
LGTM. I will remove the parameter and merge.

thommythomaso (approved):

PR #35, opened 2023-10-04
WIP: src/systems/pulpopen: Add axi-mem variant idma

thommythomaso:
* TODO: The modules should be more general. Right now, they are called 'cl' and 'fc' recalling pulp. They should be called according to what they are: one has mem interfaces only, the other axi-mem interfaces


thommythomaso:
We decided to remove the system wrappers from the iDMA repository in general. I close this PR for now.

PR #41, opened 2024-04-10
Integrate v0.6.0 into Cheshire.

yvantor:


thommythomaso:
Can you please comment on what this PR fixes? Can you also make sure that no generated files are checked in?

yvantor:
It looks that when you generate the dependencies with the `-t rtl` flag, Qestasim complains because it does not find the idma_inst64_snitch_pkg (which indeed does not appear in the compile script that Bender generates). Our first guess was the missing space in the list of [declared targets](https://github.com/pulp-platform/iDMA/blob/master/Bender.yml#L71) that should add the file to the dependencies. So far, adding that space worked.

yvantor:
@chaoqun-liang

chaoqun-liang:
> I am unsure what the whitespace changes, but the target should only be active if `rtl` and `snitch_cluster` are present. What is wrong here is that you need a stub instruction encoding for the thing to compile. I would, in your case, add the `snitch_cluster` target for now.

Yes, I would go with `snitch_cluster` for now as we already managed to get it to work this way eariler today. I can also confirm the space in between does not make a difference in this case.

thommythomaso:
I would appreciate it if you could create an issue to keep track of this. Closed.

PR #49, opened 2024-06-28
Fixes and improvements from the PULP iDMA integration effort

da-gazzi:
- register frontend now supports setting source and destination protocols
- legalizer is instantiated correctly (before it would never use bursts when any used protocol did not support it)
- register frontend correctly sets number of N-D repetitions when not using the maximum number of dimensions
- some compilation/optimization fixes (specific to `vopt`)


target/rtl/.gitignore
thommythomaso, line 1: There should be no generated files checked in.

thommythomaso:
This PR will be merged as part of #66. Closed.

PR #57, opened 2024-08-06
INIT read: Only assign `buffer_in_valid_o` if datapath req valid

micprog:


thommythomaso:
Will be merged as part of #66. Closed.

PR #60, opened 2024-09-26
src: Add multicast capabilities to Snitch DMA

colluca:
This PR implements multicast support for write transactions in the Snitch iDMA. The implementation was tested in Occamy.

The way this feature is currently implemented is not the most generic, and there may be better options. In particular, perhaps we want to consider implementing a more general mechanism to inject user-defined values in the AWUSER field.
Then it should be possible to implement multicast support simply on top of this more generic feature.


Lore0599:
This PR is closed and replaced with [PR#74](https://github.com/pulp-platform/iDMA/pull/74) which can be merged upstream

Lore0599:
This PR is closed and replaced with [PR#74](https://github.com/pulp-platform/iDMA/pull/74) which can be merged upstream

PR #96, opened 2026-05-08
idma_rt_midend: track every arbiter handshake in the choice FIFO

flaviens:
Hi!

If I understand correctly, the FIFO currently records only external bypass handshakes, but downstream responses can correspond to either external or internal requests. When the two interleave, the FIFO loses alignment and responses may be misrouted or dropped. This changes FIFO push/pop to use the unified downstream request and response handshakes, so every issued request pushes once and every returned response pops once. The bug was missed because the existing testbench ties off bypass input and does not exercise mixed traffic.

Please double check.

Thanks!
Flavien

### DIFF
--- a/src/midend/idma_rt_midend.sv
+++ b/src/midend/idma_rt_midend.sv
@@ -206,11 +206,11 @@ module idma_rt_midend #(
         .testmode_i ( 1'b0                                  ),
         .usage_o    ( /* NC */                              ),
         .data_i     ( choice                                ),
-        .valid_i    ( nd_req_valid_i & nd_req_ready_o       ),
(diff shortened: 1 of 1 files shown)


DanielKellerM:
Hi @flaviens, sorry about the noise here. We accidentally rebase-merged this PR earlier and had to force-push devel back as recovery. 

Your rt_midend fix is on our list to land. We're folding it together with PR #80 and some test work we've been carrying locally, so it will come back as a fresh PR soon. We'll cherry-pick your commit so the authorship trailer stays. If you'd rather be the PR author on the resubmit, happy to coordinate.

flaviens:
Hi! Please do the simplest. Thanks!

PR #97, opened 2026-05-08
idma_error_handler: gate WAIT_LAST_W transitions on eh_valid_i

flaviens:
Hi there!

WAIT_LAST_W is missing the eh_valid_i guard.

Because idma_eh_req_t is 1 bit (CONTINUE = 0, ABORT = 1), the idle value of eh_i is interpreted as CONTINUE when eh_valid_i is low. As a result, after a last-burst write error, the FSM immediately takes the CONTINUE path on entry to WAIT_LAST_W, so the user never gets a chance to choose ABORT.

Reproducer: with MetaFifoDepth=4, drive a write-error response with w_last_burst_i=1 and keep eh_valid_i=0. Before the fix, the FSM transitions:

IDLE -> WAIT_LAST_W -> EMIT_EXTRA_RSP -> IDLE

With the fix, it remains in WAIT_LAST_W until eh_valid_i is asserted.

Thanks!
Flavien

### DIFF
--- a/src/backend/idma_error_handler.sv
+++ b/src/backend/idma_error_handler.sv
@@ -311,29 +311,32 @@ module idma_error_handler #(
             // error happened on the last write burst of a 1D transfer. We need to emit an extra
             // response post error handling.
             WAIT_LAST_W : begin
-                // continue case (~error reporting)
-                if (eh_i == idma_pkg::CONTINUE) begin
-                    eh_ready_o = 1'b1;
-                    state_d = EMIT_EXTRA_RSP;
-                end
-                // abort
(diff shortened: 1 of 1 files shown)


DanielKellerM:
Hi @flaviens, same accidental merge story as #96.

For this one, PR #93 just landed on devel and addresses the same eh_valid gating issue, plus a couple of related guards in the surrounding states. That PR is older by a few months and has broader scope, so we kept it as the canonical fix.

You found the same bug independently though, thanks for the catch. If there's something in your version that #93 misses, let us know and we'll add it as a follow-up.

flaviens:
Sounds good!

PR #98, opened 2026-05-08
idma_channel_coupler: don't consume coupled-AW credit when popping a …

flaviens:
Hi there!

proc_credit_cnt could incorrectly consume a coupled-AW credit while popping a decoupled AW at the FIFO head. That left the later coupled AW with no saved credit and no arriving first beat, causing the AW master channel to deadlock.

This PR gates the first-consuming and credit-decrement paths on ~aw_decoupled_head, so decoupled AWs are handled only by the default pop path and never modify the coupled-AW credit counter. This is a minimal fix for master; the larger upstream decouple_rw_fix refactor remains separate.

Thanks!
Flavien

### DIFF
--- a/src/backend/idma_channel_coupler.sv
+++ b/src/backend/idma_channel_coupler.sv
@@ -127,29 +127,39 @@ module idma_channel_coupler #(
         .ready_i      ( aw_ready            )
     );
 
-    // use a credit counter to keep track of AWs to send
+    // use a credit counter to keep track of coupled AWs to send.
+    //
+    // The credit `aw_to_send_q` represents *coupled* AWs whose `first` R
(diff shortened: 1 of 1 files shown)


DanielKellerM:
Hi @flaviens, same accidental merge + force-push as #96 and #97.

We've been working on a broader stack around the channel coupler on top of PR #80, which reworks the credit accounting and fixes a handful of related bugs that we found internally. It will come in as a series of PRs once #80 is in.

Your version is a clean minimal fix and the writeup made the bug really easy to follow, so thanks for that.

flaviens:
Hi! Please do the simplest like the 2 others. Thanks!

PR #109, opened 2026-05-27
PR to merge chips-IT  changes to iDMA

RiccardoGandolfi:
Just added a parameter to the iDMA_legalizer_page_splitter module.

Related to this PR: https://github.com/pulp-platform/pulp_cluster/pull/115


yvantor:
@RiccardoGandolfi make sure you align the parameter definition to the IP name convention (i.s. `BurstLen` instead of `Burst_len`). I would also move it before type parameters since it is a pure number (so I would group it with the other `int unsigned` parameters).

RiccardoGandolfi:
> @RiccardoGandolfi make sure you align the parameter definition to the IP name convention (i.s. `BurstLen` instead of `Burst_len`). I would also move it before type parameters since it is a pure number (so I would group it with the other `int unsigned` parameters).

Should be done now

DanielKellerM (changes requested):
A few things:

1. **Rebase first**

2. **Default regression**: `BurstLen` defaults to 5, but the splitter was hardcoded to 8, so every existing variant's default burst drops 2048B→256B on a 64-bit bus (8×). The cap belongs to a specific system; keep the default at 8.

3. **These are DB values**: `'d8 = log2(max_beats_per_burst=256)` and `'d12 = log2(page_size=4096)` from `idma_axi.yml`. The legalizer already derives bursts from these per-protocol; the page splitter should take `log2(max_beats_per_burst)`/`log2(page_size)` from the DB (as `idma_legalizer_pow2_splitter` already does) instead of a hand-set param.

For the 256B bursts themselves, the reg frontend already exposes `reduce_len`/`max_llen` - set `reduce_len=1, max_llen=5`, no RTL change.

src/backend/idma_legalizer_page_splitter.sv
```diff
@@ -7,6 +7,7 @@
 
 /// Legalizer module implementing a page splitter
 module idma_legalizer_page_splitter #(
+    parameter int unsigned BurstLen     = 4'd5,
```
DanielKellerM: Default to 8 (= `log2(max_beats_per_burst)`) so existing variants keep 256-beat bursts; ideally the legalizer passes the DB value instead of a literal.

```suggestion
    parameter int unsigned BurstLen     = 4'd8,
```

RiccardoGandolfi: Fully agree, I've applied the suggestion

src/backend/tpl/idma_backend.sv.tpl
```diff
@@ -29,6 +29,8 @@ module idma_backend_${name_uniqueifier} #(
     parameter int unsigned TFLenWidth       = 32'd24,
     /// The depth of the memory system the backend is attached to
     parameter int unsigned MemSysDepth      = 32'd0,
+    /// Burst Len (for actual burst length do 8 byte * 2^(BurstLen))
+    parameter int unsigned BurstLen = 4'd5,
```
DanielKellerM:
```suggestion
    parameter int unsigned BurstLen = 4'd8,
```

DanielKellerM (commented):
Thanks for the fix + rebase. But the effective default for every generated variant is still 5. Same change needed here.

src/backend/tpl/idma_backend_synth.sv.tpl
```diff
@@ -32,6 +32,8 @@ module idma_backend_synth_${name_uniqueifier} #(
     parameter int unsigned TFLenWidth          = 32'd32,
     /// The depth of the memory system the backend is attached to
     parameter int unsigned MemSysDepth         = 32'd0,
+    /// Burst Len (for actual burst length do 8 byte * 2^(BurstLen))
+    parameter int unsigned BurstLen = 4'd5,
```
DanielKellerM:
```suggestion
    parameter int unsigned BurstLen = 4'd8,
```

src/backend/tpl/idma_legalizer.sv.tpl
```diff
@@ -21,6 +21,8 @@ module idma_legalizer_${name_uniqueifier} #(
     parameter int unsigned DataWidth       = 32'd16,
     /// Address width
     parameter int unsigned AddrWidth       = 32'd24,
+    /// Burst Len (for actual burst length do 8 byte * 2^(BurstLen))
+    parameter int unsigned BurstLen = 4'd5,
```
DanielKellerM:
```suggestion
    parameter int unsigned BurstLen = 4'd8,
```

DanielKellerM (approved):
Defaults restored to 8 across all four files.

PR #112, opened 2026-06-10
Add on-the-fly compute support with a transpose engine

DanielKellerM:
## Summary

This PR adds optional **on-the-fly compute** to the generated iDMA backends: transfers can be transformed while they stream through the DMA, with no extra memory passes. The first (and currently only) compute op is a **matrix transpose**; the architecture is an extension point for further ops.

The transpose datapath is adapted from the `datamover` (Ratha) HWPE — imported verbatim from pulp-platform/datamover@d58a985 in the first commit (original authors credited), then reworked to iDMA conventions.

## Architecture

```
opt.compute (request)  →  idma_transpose_midend  →  idma_nd_midend  →  backend
                           (tensor shape → 4-D walk)                     │
                                       write seam: idma_otf_compute ── idma_otf_transpose
```

- **Request model** (`idma_pkg`): `opt.compute = {enable, op, params}` carried per transfer; transpose params are the element mode (int8/fp16/fp32) and the tensor shape (M, N up to 4095).
- **Geometry midend**: expands a compact transpose request into the tiled `NumDim=4` ND walk (row / row-tile / col-tile) for the unmodified `idma_nd_midend`; the geometry strength-reduces to shifts except one 12x12 stride product.
- **Write-seam dispatcher**: `idma_otf_compute` latches the per-transfer options and selects one op; the AXI write manager gains an external strobe mask plus a strobe-independent beat-done so partial edge tiles drain correctly.
- **Engine**: NE x NE tile ping-pong (`NE = StrbWidth/elem`), runtime element size, element-granular edge masking. Steady state `1 + 1/NE` cycles per tile (~98% of bus peak at NE=64).

## Generation-time configuration

Compute is a **generation decision**, not an SV parameter: `IDMA_VIDMA_IDS` entries take `variant[:ops][:fd|hd]` (default `rw_axi` = transpose, full duplex). Non-listed variants render **without any compute logic**. The `hd` option builds a single tile bank — half the buffer area (`StrbWidth^2` bytes per bank, e.g. 4 KiB at 512 bit) for roughly half the streaming rate.

Constraints, enforced at generation/elaboration time: compute variants need a single AXI write port and `NO_ERROR_HANDLING`; sources/destinations must be readable/writable up to the tile-padded bounds (documented contract; writes are strobe-masked, reads are not).

## Verification

- Standalone engine regression against a DPI-C golden model, in both duplex modes
[...]

PR #113, opened 2026-06-10
inst64: Drive on-the-fly transpose; add the snitch integration harness

DanielKellerM:
## Summary

Drives on-the-fly transpose through the inst64 (Snitch) frontend and adds a standalone Snitch integration harness.

inst64's backend (`rw_axi_rw_init_rw_obi`) is multi-write and therefore cannot host the #112 single-AXI-write transpose engine. Instead, the transpose is executed by **address generation**: `idma_transpose_midend` (new `AddrGenTranspose` mode) emits an element-granular `NumDim=3` swapped-stride ND program (`out_T[c][r] = in[r][c]`) and clears `opt.compute`, so the backend runs a plain strided copy — no engine, no tile buffer. OBI/TCDM is random-access, so this is the natural (and only needed) mechanism for inst64; the #112 FF engine remains the full-throughput path for AXI↔AXI transpose on `rw_axi`.

## What's here

- **Frontend** (`idma_inst64_top`): decode the transpose DMCPY into `opt.compute`, splice `idma_transpose_midend` ahead of `idma_nd_midend`, and reject malformed requests (no hardware / reserved mode / zero dim / unaligned dst).
- **Address-gen** (`idma_transpose_midend`, `AddrGenTranspose`): swapped-stride transpose with an optional `BankSkew` that pads the dst row pitch by one bus-word when the column stride would hammer a single TCDM bank — makes the per-column word stride odd, so writes round-robin across all banks on any power-of-2-bank L1, at ≤1 word/row cost.
- **Harness** (`systems/snitch`): standalone BFM driving the accelerator port plus AXI and native-OBI sim memories.

## Testing

The self-checking harness TB sweeps a geometry list in one elaboration (int8/fp16/fp32, square/rectangular/odd) over the real OBI/TCDM port:

- **OBI→OBI** — transpose a tile within L1/TCDM
- **AXI→OBI** — load an external matrix into TCDM transposed
- back-to-back (consecutive cases), cross-transfer compute-leak (a plain copy after transposes must not inherit `opt.compute`), and malformed-request reject
- both `BankSkew` off (contiguous N×M) and on (padded N×M′, padding stays sentinel)

A standalone copy harness (`tb_idma_inst64_copy`) also passes. All changes live in `src/` — no codegen/templatization.

## Notes

- Address-gen is element-granular (M·N transactions): ideal on random-access OBI/TCDM, slower than a burst on AXI. The FF engine (#112) stays the full-throughput path for AXI↔AXI on `rw_axi`.
- Rebased on current `devel`; the earlier transpose-engine commits are dropped (they landed via #112).

### DIFF
--- a/Bender.yml
+++ b/Bender.yml
@@ -97,6 +97,18 @@ sources:
       # Level 2
       - src/frontend/desc64/idma_desc64_top.sv
 
+  # Snitch inst64 standalone harness; needs snitch_cluster for idma_inst64_top + opcode pkg
+  - target: all(snitch_cluster, idma_test)
+    files:
+      # Level 0
+      - systems/snitch/test/idma_inst64_tb_pkg.sv
+      # Level 1
+      - systems/snitch/test/idma_inst64_drv_if.sv
+      - systems/snitch/test/idma_inst64_base.sv
+      # Level 2
+      - systems/snitch/test/tb_idma_inst64_copy.sv
+      - systems/snitch/test/tb_idma_inst64_transpose.sv
+
   # Synthesis wrappers
   - target: synth
     files:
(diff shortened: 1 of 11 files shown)


DanielKellerM:
Superseded by #141.

PR #120, opened 2026-06-11
ci: Cache the bender git database keyed on the lockfile

DanielKellerM:
- `BENDER_DB_DIR` relocates bender's bare-repo database to a cacheable path; `actions/cache` keys it on `Bender.lock` with a restore-key fallback, so a stale cache self-heals via incremental fetch instead of a full re-clone.
- Warm checkouts need no network (~0.09 s vs ~2 s cold locally; cache is ~18 MB), removing github.com clone flakiness from build/docs/deploy.
- The database (bare repos) is cached rather than the project-local `.bender/`: it is smaller (18 MB vs 26 MB), position-independent, and reusable across lock bumps, while `.bender/` checkouts carry absolute-path alternates into the database and are rebuilt from it in 0.08 s anyway.

Note: a committed `.bender.yml` setting `db_dir` would silently override the env var — none exists today.

### DIFF
--- a/.github/actions/bender-db-cache/action.yml
+++ b/.github/actions/bender-db-cache/action.yml
@@ -1,0 +1,18 @@
+# Copyright 2026 ETH Zurich and University of Bologna.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+name: Cache Bender database
+description: Relocate bender's bare-repo git database to a cached path
+# A committed .bender.yml setting db_dir would silently override BENDER_DB_DIR.
+runs:
+  using: composite
+  steps:
+    - shell: bash
+      run: echo "BENDER_DB_DIR=$HOME/.cache/bender-db" >> "$GITHUB_ENV"
+    - uses: actions/cache@v4
+      with:
+        path: ~/.cache/bender-db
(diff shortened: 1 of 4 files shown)


.github/actions/bender-db-cache/action.yml
```diff
@@ -0,0 +1,18 @@
+# Copyright 2026 ETH Zurich and University of Bologna.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+name: Cache Bender database
+description: Relocate bender's bare-repo git database to a cached path
+# A committed .bender.yml setting db_dir would silently override BENDER_DB_DIR.
```
DanielKellerM: remove verbose comment

.github/actions/bender-db-cache/action.yml
```diff
@@ -0,0 +1,18 @@
+# Copyright 2026 ETH Zurich and University of Bologna.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
```
DanielKellerM, line 4: missing author line.

DanielKellerM (commented):
nitpick

PR #123, opened 2026-06-12
backend: Fix multi-head datapath bugs and add a verification testbench

DanielKellerM:
The multi-head capability (#85) was merged but **never elaborated** — the only backend testbench pins both heads to 0, so a multi-head variant was never built or simulated. Generating and compiling a 2-read-head + 1-write-head AXI backend (`2r_axi_w_axi`) surfaced several real bugs in the generated transport layer and backend:

- write datapath `dst_head` was sourced from `src_head`
- write/aw head gating dereferenced `dst_head` on the **read** datapath request
- per-head read/write instances drove the whole master-port array instead of their `[head]` slice (missing the per-head bus index)
- the write-response head FIFO drove an undeclared signal
- a single-protocol multi-head read was treated as a single (untagged) port → meta-channel type mismatch
- the per-head read buffer packed dimensions were inverted (32 vs 16-bit)
- the AR meta valid gated on the read-datapath head, not the AR head

## Verification

Adds `test/tb_idma_backend_multihead.sv` — a directed, self-checking testbench for the 2-read-head/1-write-head AXI backend that gives **each read head its own memory** and verifies `src_head` routing (head 0 reads its memory, head 1 reads its own). It reuses the per-head-memory harness pattern from the vidma inst64 verification harness. Result in Questa: `ALL CHECKS PASSED`.

## Impact

Existing single-head variants are unaffected: the only change to their generated RTL is the `dst_head` fix, which is a no-op at head 0. Verified by diffing generated RTL before/after across all default variants. The new testbench is behind a dedicated `multihead` Bender target and is exercised only by the internal CI multi-head job (off by default in public builds).

## Test scope / caveats (independently verified)

The two testbenches instantiate the backend with `HardwareLegalizer=1` (the recommended/default config), so they exercise the **hardware-legalizer datapath**. Independent reproduction confirms head routing is genuinely checked (cross-head transfers read one head's preloaded memory and verify the *other* head's memory).

Two honest limitations:
- The `idma_backend.sv.tpl` `dst_head ← src_head` one-liner lives in the **SW-legalizer / bypass** path (`gen_no_hw_legalizer`), which these TBs do **not** exercise — with `HardwareLegalizer=1`, `w_dp_req` comes from the legalizer (already correct). The fix is correct by inspection but is **not** covered by these TBs.
- The multi-head **bypass** path (`HardwareLegalizer=0`) appears to have separate issues (a directed run there does not reach a pass) and is left as a **follow-up** — out of scope for this bring-up, which targets the default legalizer config.

All other fixes (per-head bus index, AR/AW/`w_dp` head gating, read-buffer packing, write-response head FIFO) are exercised and necessary for elaboration + routing.

### DIFF
--- a/Bender.yml
+++ b/Bender.yml
@@ -131,7 +131,6 @@ sources:
       - target/rtl/tb_idma_generated.sv
 
   # Multi-head directed backend testbenches
-  # (need IDMA_ADD_IDS="2r_axi_w_axi 2rw_axi" at gen time)
   - target: multihead
     files:
       - test/tb_idma_backend_multihead.sv
--- a/Bender.yml
+++ b/Bender.yml
@@ -129,3 +129,8 @@ sources:
   - target: idma_test
     files:
       - target/rtl/tb_idma_generated.sv
+
+  # Multi-head directed backend testbench (needs IDMA_ADD_IDS=2r_axi_w_axi at gen time)
+  - target: multihead
+    files:
+      - test/tb_idma_backend_multihead.sv
--- a/Bender.yml
+++ b/Bender.yml
@@ -130,7 +130,9 @@ sources:
     files:
       - target/rtl/tb_idma_generated.sv
 
-  # Multi-head directed backend testbench (needs IDMA_ADD_IDS=2r_axi_w_axi at gen time)
+  # Multi-head directed backend testbenches
+  # (need IDMA_ADD_IDS="2r_axi_w_axi 2rw_axi" at gen time)
   - target: multihead
     files:
       - test/tb_idma_backend_multihead.sv
+      - test/tb_idma_backend_multihead_rw.sv
--- a/src/backend/tpl/idma_transport_layer.sv.tpl
+++ b/src/backend/tpl/idma_transport_layer.sv.tpl
@@ -217,7 +217,7 @@ _rsp_t ${mh_format['aw'][protocol]}${protocol}_write_rsp_i,
     // shifted data flowing into the buffer
 % if not one_read_port:
     % for p in used_read_protocols:
-    byte_t [StrbWidth-1:0] ${mh_format['ar'][p]}${p}_buffer_in;
+    byte_t ${mh_format['ar'][p]}[StrbWidth-1:0] ${p}_buffer_in;
     % endfor
 % endif
     byte_t [StrbWidth-1:0] buffer_in;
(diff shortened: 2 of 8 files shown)


src/backend/tpl/idma_transport_layer.sv.tpl
```diff
@@ -487,7 +487,7 @@ ${rendered_write_ports[write_port]}
         .data_i     ( w_dp_req_i.dst_head                            ),
         .valid_i    ( w_resp_fifo_in_valid && w_resp_fifo_in_ready   ),
         .ready_o    ( /* NOT CONNECTED */                            ),
-        .data_o     ( w_resp_fifo_out_multihead                      ),
+        .data_o     ( w_resp_fifo_out_head                      ),
         .valid_o    ( /* NOT CONNECTED */                            ),
```
DanielKellerM, line 491: verify this is correct

src/backend/tpl/idma_transport_layer.sv.tpl
```diff
@@ -517,9 +517,9 @@ ${rendered_write_ports[write_port]}
                 w_dp_rsp_mux       = ${wp}_w_dp_rsp;
                 ${wp}_w_dp_rsp_ready = w_dp_rsp_mux_ready;
     % else:
-                w_dp_rsp_mux_valid = ${wp}_w_dp_rsp_valid [w_resp_fifo_out_multihead];
```
DanielKellerM, line 520: verify this is correct

util/mario/transport_layer.py
```diff
@@ -106,7 +106,7 @@ def render_read_mgr_inst(prot_id: str, prot_ids: dict, db: dict) -> dict:
                 else:
                     read_meta_valid = f'''\
 (ar_req_i.src_protocol == idma_pkg::{db[rp]["protocol_enum"]}) & \
-(r_dp_req_i.src_head == {curr_head}) & \
+(ar_req_i.src_head == {curr_head}) & \
 ar_valid_i\
```
DanielKellerM, line 110: verify this is correct

Bender.yml
```diff
@@ -129,3 +129,10 @@ sources:
   - target: idma_test
     files:
       - target/rtl/tb_idma_generated.sv
+
+  # Multi-head directed backend testbenches
+  # (need IDMA_ADD_IDS="2r_axi_w_axi 2rw_axi" at gen time)
```
DanielKellerM: remove verbose comment

PR #119, opened 2026-06-11
docs: Align Starlight site with generated hierarchy graphs

DanielKellerM:
**Draft — work in progress.** A single workspace to bring the new Starlight docs site (#103) and the regenerated module-hierarchy graphs (#117) into alignment, and to refresh the docs to the current state of the repo.

### Starter in this PR
- New `architecture/hierarchy.md` page wired to the `ast2dot`-generated graphs (`/fig/graph/*.png`).
- `make idma_doc_site` target that copies the generated `doc/fig/graph/*.png` into the site's `public/fig/graph/`.
- `.gitignore` so the generated graphs aren't committed.

### TODO (to be filled in)
- [ ] Regenerate the hierarchy graphs from the current RTL (`make idma_doc_all`) and verify they render in the site.
- [ ] Bring the architecture/guide pages up to date with the latest repo (post pickle/uv/rename changes).
- [ ] Decide whether/how to build + deploy the Starlight site in CI (currently Pages serves the Sphinx docs).
- [ ] Reconcile the static SVGs (`public/fig/*.svg`) vs the generated graphs.

Preview of the current site: https://danielkellerm.github.io/iDMA/ (fork).

PR #128, opened 2026-06-16
build: Self-bootstrap the uv environment for make idma_hw_all

DanielKellerM:
## What

`make idma_hw_all` (and the other generators) now provision their own Python environment. If Mako/peakrdl are not importable, a local `.venv` is created via `uv sync --locked` and prepended to PATH. An already-activated venv or `uv run make …` (as in CI) is detected and left untouched, so existing flows are unchanged.

## Why

Downstream SoCs that pull iDMA via Bender (e.g. pulp_cluster) had to replicate the env setup — `uv sync` + `source .venv/bin/activate` — before calling `make -C $(bender path idma) idma_hw_all`. With this, the low-level handling lives in the IP and a consumer just calls:

```make
generate_idma_rtl:
	$(MAKE) -C $(shell bender path idma) idma_hw_all
```

## Notes

- Requires `uv` on PATH when cold; otherwise it errors with the install hint (no silent `curl | sh`). uv is already a documented prerequisite.
- Tested locally: cold (no venv, mako-less python) provisions and generates all variants; idempotent re-run reuses the `.venv`; `uv run make idma_hw_all` unchanged.

Docs follow-up: simplify the quickstart build snippet (#119) to the bare `make idma_hw_all` once this lands.

### DIFF
--- a/idma.mk
+++ b/idma.mk
@@ -17,6 +17,24 @@ VERILATOR   ?= verilator
 VLOGAN      ?= vlogan
 VSIM        ?= vsim
 
+# Provision a local uv venv when the generator deps are not already available.
+IDMA_VENV_PY := $(IDMA_ROOT)/.venv/bin/python
+ifeq ($(shell $(PYTHON) -c 'import mako' >/dev/null 2>&1 && echo ok),ok)
+else ifeq ($(shell $(IDMA_VENV_PY) -c 'import mako' >/dev/null 2>&1 && echo ok),ok)
+  PYTHON      := $(IDMA_VENV_PY)
+  export PATH := $(IDMA_ROOT)/.venv/bin:$(PATH)
+else ifeq ($(shell command -v uv >/dev/null 2>&1 && echo ok),ok)
+  $(info iDMA: provisioning the generator environment (uv sync --locked) ...)
+  _idma_uv_sync := $(shell cd $(IDMA_ROOT) && uv sync --locked 1>&2 || echo FAIL)
+  ifeq ($(_idma_uv_sync),FAIL)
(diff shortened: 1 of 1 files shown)


idma.mk
```diff
@@ -17,6 +17,27 @@ VERILATOR   ?= verilator
 VLOGAN      ?= vlogan
 VSIM        ?= vsim
 
+# --- Self-bootstrapping generator environment (uv) --------------------------
+# `make idma_hw_all` (and the other generators) work with no pre-activated venv:
+# if Mako/peakrdl are not already importable, a local .venv is provisioned via
+# `uv sync --locked` and prepended to PATH. An activated venv or `uv run make`
+# (as in CI) is detected and left untouched. Requires `uv` on PATH when cold.
```
DanielKellerM: remove verbose comments

idma.mk
```diff
@@ -17,6 +17,27 @@ VERILATOR   ?= verilator
...
+IDMA_VENV_PY := $(IDMA_ROOT)/.venv/bin/python
+ifeq ($(shell $(PYTHON) -c 'import mako' >/dev/null 2>&1 && echo ok),ok)
+  # default interpreter already has the deps (activated venv / uv run / CI)
```
DanielKellerM: remove

idma.mk
```diff
@@ -17,6 +17,27 @@ VERILATOR   ?= verilator
...
+else ifeq ($(shell $(IDMA_VENV_PY) -c 'import mako' >/dev/null 2>&1 && echo ok),ok)
+  # local .venv already provisioned
```
DanielKellerM: remove

PR #129, opened 2026-06-16
release: iDMA 0.7.0

DanielKellerM:
> **Draft release PR — iDMA 0.7.0** (`devel` → `master`). Consolidates the 85 commits on `devel` since 0.6.5. 0.7.0 is a **breaking** minor (0.x minor = breaking axis).

## Breaking changes

Seven, and a downstream integrator needs all of them:

- **Register blocks regenerated from SystemRDL (PeakRDL)**, replacing the reggen/`reg_bus` flow ([#73](https://github.com/pulp-platform/iDMA/pull/73)).
- **The reg-frontend config bus is CPUIF-selectable**, `apb4-flat` by default; `register_interface` is dropped as a dependency ([#148](https://github.com/pulp-platform/iDMA/pull/148)). Breaking for `reg_bus` integrators.
- **Multi-head DMA** adds `idma_pkg` types and selects channels per transfer via `src_head`/`dst_head` ([#85](https://github.com/pulp-platform/iDMA/pull/85), [#123](https://github.com/pulp-platform/iDMA/pull/123)).
- **The tracer is generated one header per backend id** ([#186](https://github.com/pulp-platform/iDMA/pull/186)). `` `include "idma/tracer.svh" `` becomes `` `include "idma/tracer_<id>.svh" `` for the variant being traced; the shared header keeps only the id-independent helpers.
- **Trace signal keys are qualified by direction** ([#186](https://github.com/pulp-platform/iDMA/pull/186)). A consumer reading `axi_rsp_ready` now reads `axi_read_rsp_ready`. Without this, a protocol present on both sides (INIT) emitted the same key twice and the read channel was silently dropped from the trace.
- **`common_cells` v2 is required** ([#99](https://github.com/pulp-platform/iDMA/pull/99), [#180](https://github.com/pulp-platform/iDMA/pull/180)). Ecosystem-wide bump.
- **The pickle moves to `target/pickle`** ([#110](https://github.com/pulp-platform/iDMA/pull/110), [#114](https://github.com/pulp-platform/iDMA/pull/114)), replacing morty.

## Added

[...]

PR #131, opened 2026-06-16
frontend: Expose the reg frontend as a native APB slave

DanielKellerM:
## What

The register-based frontend (`idma_reg32_*`/`idma_reg64_*`) now exposes a **native APB4 slave** instead of a register_interface reg_bus slave. The `reg_to_apb` shim is deleted and `register_interface` is dropped from `Bender.yml`.

## Why

Since the SystemRDL/PeakRDL migration (#73), the generated `idma_*_reg_top` is **already an APB4 block** (`--cpuif apb4-flat`). The reg frontend wrapped that APB block behind a reg_bus port + a `reg_to_apb` shim purely to keep the old port type — i.e. it converted reg_bus → APB to reach a register file that is APB anyway. The desc64 frontend is already native-APB; this brings the reg frontend in line so **every iDMA register interface is PeakRDL-native APB**, with no register_interface idiom left.

`reg_to_apb` was the *only* consumer of `register_interface` in the whole repo, so the dependency goes away entirely (no transitive fallout — nothing else pulls it).

## How the launch backpressure is preserved

Reading `next_id` launches a transfer, and the read must stall until the backend accepts it. Previously this rode on the reg_bus `.ready` field (via a `read_happens ? arb_ready : ...` trick on the shim's response). `next_id` is an `external` SystemRDL register, so the PeakRDL APB CPUIF holds `s_apb_pready` low until `hwif_in.next_id.rd_ack` is asserted. The stall now lives on that ack:

```systemverilog
assign dma_hw2reg[i].next_id[c].rd_ack = arb_ready[i];
```

The old `proc_dma_backpressure` block (which also had a latent struct-into-scalar assignment in its false branch) is removed.

## Breaking change

This is a **port-level breaking change** for integrators wiring the reg frontend over reg_bus — they now connect APB (`apb_req_t`/`apb_rsp_t`, define via `APB_TYPEDEF_ALL`) or place their own bridge. Appropriate for the breaking 0.7.0 line; the integration guide is updated. APB-native SoCs (e.g. pulp_cluster) connect straight through.

## Verification

- `make idma_hw_all` regenerates all four reg frontends clean; no residual `reg_to_apb`/`reg_req_t`.
- Full-design elaboration with `slang` over `bender script flist-plus` (whole iDMA + deps + generated reg32_3d frontend driven by APB types) passes with **0 errors**.
- `register_interface` no longer appears in `Bender.lock`.

Part of the iDMA 0.7.0 release (#129).

### DIFF
--- a/Bender.lock
+++ b/Bender.lock
@@ -44,16 +44,6 @@ packages:
     dependencies:
     - common_cells
     - common_verification
-  register_interface:
-    revision: d6e1d4cdaab7870f4faf3f88a1c788eaf5ac129d
-    version: 0.4.7
-    source:
-      Git: https://github.com/pulp-platform/register_interface.git
-    dependencies:
-    - apb
-    - axi
-    - common_cells
-    - common_verification
   tech_cells_generic:
     revision: 7968dd6e6180df2c644636bc6d2908a49f2190cf
     version: 0.2.13
--- a/Bender.yml
+++ b/Bender.yml
@@ -18,7 +18,6 @@ dependencies:
   axi_stream:          { git: "https://github.com/pulp-platform/axi_stream.git",          version: 0.1.1  }
   common_cells:        { git: "https://github.com/pulp-platform/common_cells.git",        version: 1.39.0 }
   common_verification: { git: "https://github.com/pulp-platform/common_verification.git", version: 0.2.5  }
-  register_interface:  { git: "https://github.com/pulp-platform/register_interface.git",  version: 0.4.7  }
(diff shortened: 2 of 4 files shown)


DanielKellerM:
Superseded by #148 (selectable reg-frontend CPUIF), now merged to devel. #148's default apb4-flat reproduces this PR exactly and adds obi-flat / axi4-lite-flat; both drop the reg_to_apb shim + register_interface. Closing in favor of #148.

PR #135, opened 2026-06-18
frontend: Model status/done_id as internal status registers

DanielKellerM:
Refines the external-register read-ack fix (#134) by modeling the two observational registers correctly instead of gating their ack.

\`status\` (busy) and \`done_id\` have **no read side-effect**, so they need not be \`external\`. This drops \`external\` on them in the RDL and drives the registered value through the PeakRDL \`hwif .next\` input; PeakRDL then emits plain internal storage with **no external read handshake and no \`assert_bad_ext_rd_ack\`** for those registers. \`next_id\` keeps its external read-launch interface with the \`arb_ready\`-gated ack from #134 (its read launches a transfer and must back-pressure).

**Net effect:** the external read handshake/assertion now governs only \`next_id\` (the one register that genuinely needs it); the change removes logic rather than gating it.

**Trade-off:** \`status\`/\`done_id\` reads are now registered (≤1-cycle-delayed) rather than live-combinational. This is benign for busy/done polling (you never read a value that was not true, just up to a cycle late) and the **SW-facing register map is unchanged** (offsets, width, \`sw=r\` access identical).

**Status:** verified structurally — regenerates cleanly, the generated reg block elaborates (slang, 0 errors), and \`external_rd_ack\` now ORs only \`next_id\`. Draft pending the same cluster/UVM sim validation #134 had (assertion clears + a SW polling read returns the expected registered value).

### DIFF
--- a/src/frontend/reg/tpl/idma_reg.sv.tpl
+++ b/src/frontend/reg/tpl/idma_reg.sv.tpl
@@ -203,27 +203,24 @@ module idma_${identifier} #(
     end
 
     // observational registers
+    // status and done_id are internal hw=w regs (no read side-effect): drive the
+    // registered value. next_id is external (read launches a transfer): ack only
+    // on request, held until the arbiter accepts (read back-pressure).
     for (genvar c = 0; c < NumStreams; c++) begin : gen_hw2reg_connections
-        assign dma_hw2reg[i].status[c].rd_data.busy  = {midend_busy_i[c], busy_i[c]};
-        assign dma_hw2reg[i].status[c].rd_ack = dma_reg2hw[i].status[c].req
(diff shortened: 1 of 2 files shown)


src/frontend/reg/tpl/idma_reg.sv.tpl
```diff
@@ -203,27 +203,24 @@ module idma_${identifier} #(
     end
 
     // observational registers
+    // status and done_id are internal hw=w regs (no read side-effect): drive the
+    // registered value. next_id is external (read launches a transfer): ack only
+    // on request, held until the arbiter accepts (read back-pressure).
```
DanielKellerM, line 208: remove verbose comments

DanielKellerM (commented):
fix

PR #140, opened 2026-06-22
test: Drive OBI backend ports with native obi_sim_mem

DanielKellerM:
## Problem
The `vsim-sim-random` OBI tests (`rw_obi`, `r_obi_w_axi`, `r_axi_w_obi`, `snitch_*`) hang on rw-decoupled transfers ≳512 B. Root cause is in the **testbench**, not the iDMA RTL: the OBI path converts each backend OBI port to AXI via the vendored `per2axi` peripheral converter, joins read+write with `axi_rw_join`, and shares one `axi_sim_mem`. Under rw-decoupled burst traffic the per2axi bridge stops accepting memory responses (`b_ready=0`/`r_ready=0` while the memory holds `b_valid`/`r_valid`), so nothing completes. The backend's native OBI write is independently proven deadlock-free against a responsive memory.

## Fix
Drive the backend's OBI ports with a native `obi_sim_mem` (separate read/write memories; scoreboard seeds src in the read mem and checks dst in the write mem). Drop the per2axi bridges, `axi_rw_join`, and AXI throttle/multicut for OBI; repoint the OBI highlighters and watchdog to the native OBI handshakes. AXI path unchanged.

## Validation
Locally (Questa): rw_obi (incl. the previously-hanging seed), r_obi_w_axi, r_axi_w_obi, snitch variants, and rw_axi (AXI path) suites all green. This PR runs from an upstream branch so the GitLab sims execute in CI (fork PRs skip them).

### DIFF
--- a/test/include/tb_tasks.svh
+++ b/test/include/tb_tasks.svh
@@ -82,15 +82,15 @@
 `endif
 
 `ifdef PROT_OBI
-    // write a byte to the OBI AXI-attached memory
+    // seed a byte into the OBI source (read) memory
     task write_byte_obi_axi_mem (
         input byte_t byte_i,
         input addr_t addr_i
     );
-        i_obi_axi_sim_mem.mem[addr_i] = byte_i;
+        i_obi_read_sim_mem.mem[addr_i] = byte_i;
     endtask
 
-    // read a byte from the OBI AXI-attached memory
+    // read a byte back from the OBI destination (write) memory
     task read_byte_obi_axi_mem (
         output byte_t byte_o,
         input  addr_t addr_i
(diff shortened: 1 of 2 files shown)


DanielKellerM:
Rebased onto current devel and fixed the multiverse elaboration failure (`vopt-7063`).

The native `obi_sim_mem` backdoor refs in `tb_tasks.svh` (`write_byte_obi_axi_mem`→`i_obi_read_sim_mem`, `read_byte_obi_axi_mem`→`i_obi_axi_sim_mem`) compile under `PROT_OBI`, which is shared across the whole compile unit — so they must *resolve* for every backend, not just OBI ones. The read/write OBI mems were only instantiated for OBI read/write respectively, so non-OBI (`rw_axi_rw_axis`) and single-direction-OBI (`r_obi_w_axi`, `r_axi_w_obi`) backends couldn't elaborate.

Fix: declare per-direction `dummy_mem_t` placeholders (real mem where the OBI direction is used, dummy otherwise) — the same pattern already used for fully-unused protocols. The dummies satisfy elaboration only; the OBI tasks are runtime-guarded by `used_protocols`, so they're never accessed on a dummy.

Validated locally (Questa 2023.4): all OBI backend combos pass simple + medium (7500 jobs / 404 kiB), including full-OBI `rw_obi` and `rw_axi_rw_init_rw_obi`.

PR #147, opened 2026-07-03
frontend: Eliminate external_pending launch deadlock (non-external next_id via rd_swacc)

DanielKellerM:
## Problem

Reading `next_id` launches a DMA. Since #73 the register frontend models `next_id`/`status`/`done_id` as PeakRDL `external` registers. peakrdl-regblock emits an `external_pending` latch driving `cpuif_req_stall_rd/wr` — a **whole register-block stall** while any external access is outstanding.

#134 then gated the external `rd_ack` on the decoded `req` strobe. But `external_pending` *masks* `cpuif_req`, so `req` is high for only one cycle; once `external_pending` latches, `req` drops to 0. If the arbiter grant (`arb_ready`) is not high in that first cycle, `rd_ack = req & arb_ready` can never fire again → `external_pending` is stuck high → the whole reg block is frozen. When the config port shares interconnect ordering with other traffic (e.g. a mem-tile DMA whose config slave shares a demux/NoC path with CPU stores), this frozen block back-pressures and **deadlocks unrelated memory traffic**.

The pre-0.7.0 (reggen) frontend never had this: `next_id` was `hwext + hwre` (a hardware-external register with a `.re` read strobe) — no `external_pending`, no whole-block stall; the read stalled only locally via the wrapper's reg-bus `ready`.

## Fix

Model the observation registers as **non-`external`**, reproducing the reggen behaviour in PeakRDL:
- `next_id`: a normal internal register (`sw=r; hw=w`) with a **`rd_swacc` read-strobe** (declared as a UDP) driving the launch — a combinational one-cycle pulse, exactly like the old `.re`.
- `status`/`done_id`: internal (from #135).

With **no** `external` register, peakrdl emits `external_pending = 0` and hardwires `cpuif_req_stall = '0` — no whole-block stall, no deadlock, and no `assert_bad_ext_rd_ack`. The frontend fires the launch off `rd_swacc`; no ack handshake, no hold latch.

## Relation to other PRs

This branch incorporates:
- **#131** (native-APB frontend) — the APB reg block interface.
- **#135** (internal `status`/`done_id`) — required, since *all* observation regs must be non-external to fully remove `external_pending`.

and fixes **#134**'s launch-read deadlock via the `next_id` → `rd_swacc` change.

Suggested merge order: **#131 → #135 → this**. It can also be merged standalone (it contains all three); if #131/#135 land first, this rebases down to just the `next_id` → `rd_swacc` change + the `rd_swacc` UDP declaration.

## Test

Validated on a Gwaihir mem-tile DMA (l2_dma L2↔L2 passthrough), where the config port shares the mem-tile demux/NoC path with CVA6 SPM stores. On `devel` + #134 this hard-deadlocks the SoC; with this fix it runs to completion (SLINK SUCCESS), matching the 0.6.5 baseline, with zero assertions.

### DIFF
--- a/src/frontend/reg/idma_reg.rdl
+++ b/src/frontend/reg/idma_reg.rdl
@@ -9,6 +9,11 @@
 `ifndef IDMA_REG_REG_RDL
 `define IDMA_REG_REG_RDL
 
+property rd_swacc {
+    type = boolean;
+    component = field;
+};
+
 addrmap idma_reg #(
     longint unsigned SysAddrWidth = 32, // Address width
     longint unsigned NumDims = 2, // Number of dimensions available
@@ -68,9 +73,10 @@ addrmap idma_reg #(
         name = "next_id";
         desc = "Next ID, launches transfer, returns 0 if transfer not set up properly.";
         default sw = r;
-        default hw = rw;
+        default hw = w;
         field {
             desc = "Next ID, launches transfer, returns 0 if transfer not set up properly.";
+            rd_swacc = true;
         } next_id [31:0] = 0;
     };
 
@@ -153,9 +159,9 @@ addrmap idma_reg #(
     };
 
              conf     conf;
-    external status   status[16];
-    external next_id  next_id[16];
-    external done_id  done_id[16];
+             status   status[16];
+             next_id  next_id[16];
+             done_id  done_id[16];
              dst_addr dst_addr[SysAddrWidth/32] @ 0xD0;
              src_addr src_addr[SysAddrWidth/32];
              length   length[SysAddrWidth/32];
--- a/src/frontend/reg/tpl/idma_reg.sv.tpl
+++ b/src/frontend/reg/tpl/idma_reg.sv.tpl
@@ -6,8 +6,6 @@
 // - Michael Rogenmoser <michaero@iis.ee.ethz.ch>
 // - Thomas Benz <tbenz@iis.ee.ethz.ch>
 
-`include "apb/typedef.svh"
-
 /// Description: Register-based front-end for iDMA
 module idma_${identifier} #(
   /// Number of configuration register ports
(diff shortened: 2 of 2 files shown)


DanielKellerM:
## SW C-header note (for downstream consumers)

Since this migrates the reg frontend fully to SystemRDL, downstream SW that consumed the old reggen `--cdefines` header (`idma_reg64_2d.hjson`) needs the equivalent from PeakRDL. Two data points:

- **cheshire** handles this in [pulp-platform/cheshire#281](https://github.com/pulp-platform/cheshire/pull/281): it generates `regs/idma.h` directly from this RDL with
  `peakrdl c-header src/frontend/reg/idma_reg.rdl -b ltoh --type-style hier --rename idma_reg64_2d -P SysAddrWidth=64 -P NumDims=2 -P Log2NumDims=1`
  and switches its `dma.h` to `offsetof(idma_reg64_2d_t, …)`. The moved 2D-dim offsets (`0x100`) then fall out of the generated struct layout automatically.
- **Validated** end-to-end on a Gwaihir mem-tile DMA (config port sharing the mem-tile demux/NoC path with CPU stores): with this frontend fix + the generated struct header, `l2_dma` (1D + 2D phases) runs to SLINK SUCCESS with zero assertions.

Flagging per @-review discussion: it may be worth having **iDMA itself ship a `c-header` make target** (e.g. in `idma.mk`, alongside the existing `regblock`/`raw-header` rules) so every consumer doesn't re-derive the `peakrdl c-header` invocation. Not required for this PR — #281 shows it works from the RDL directly — but it would centralize the naming/param choices.

DanielKellerM:
Superseded by #153 (frontend/reg-deadlock-fix): a minimal, template-only fix that keeps next_id external and preserves the canonical BLOCKING launch contract, fixing the external_pending deadlock at the rd_ack driver with a pending latch — plus the first directed reg-frontend testbench proving it (non-vacuous: the watchdog fires on the unfixed template). This #147 approach (non-external next_id + rd_swacc) changed the contract to poll-based, could drop a launch under contention, and was untested. Closing in favor of #153.

PR #149, opened 2026-07-06
[CI] Modifications to CI

da-gazzi:
Slight modifications to the CI:

- don't run `deploy` job on pull requests - instead, run it when pushing to `devel` or `master` (or manually)
- clean up the `deploy.py` script


DanielKellerM (approved):
LGTM. In a follow up I'll add `rw_axi_rw_init_rw_obi` also as default in `idma.mk` to match deploy branch and make idma_hw_all targets

PR #150, opened 2026-07-06
DRAFT: Generate C headers and fix small RTL bug in register file

da-gazzi:
- Generate C headers in `target/sw` with both the `raw-header` and `c-header` PeakRDL plugins
- Fix the assignment of `dma_ctrl_rsp_o.ready` in the register file top-level template
- fix arbitration of `stream_idx` in the register frontend

Based on the CI PR; should be merged after that


da-gazzi:
closing due to diverging history; I will open a new PR for the C headers.

PR #152, opened 2026-07-08
hw: Fix illegal obi response assign in `idma` top

gbellocchi:
Found error from previous PR #146 , which is failing synthesis at elaboration-time.


DanielKellerM (approved):
Thx. LGTM

PR #153, opened 2026-07-08
frontend: Fix reg next_id launch deadlock

DanielKellerM:
## What

Fixes a launch deadlock in the register frontend's `next_id` read — introduced by the SystemRDL/PeakRDL migration (#73) and inherited by devel/#148 — and adds the first directed reg-frontend testbench.

## Root cause

`next_id` is a PeakRDL `external` register: reading it launches a transfer and the read stalls (blocking) until the backend arbiter grants (`arb_ready`). When the launch isn't granted in its single active cycle, the reg_top latches `external_pending` and **masks its own CPUIF req** — so `next_id.req` becomes a one-cycle pulse. The #134 gate `rd_ack = req & ~req_is_wr & arb_ready` still references that maskable `req`, so once `req` drops, `rd_ack` can never fire, `external_pending` never clears, and the whole reg block freezes.

## Fix

Reconstruct the canonical (pre-#73 reggen) *held* read-strobe with a per-`[reg][stream]` pending latch, entirely in the CPUIF-agnostic launch layer:

- `nxt_read_seen` = the single-cycle req pulse; `nxt_read_pending_q` sets on `seen & ~arb_ready`, clears on `pending & arb_ready`
- `arb_valid = |(seen | pending)` — holds the arbiter request across the stall (satisfies `rr_arb_tree` LockIn)
- `rd_ack = (seen | pending) & arb_ready` — driven by the latch, not the maskable req

`next_id`/`status`/`done_id` stay `external` (no RDL change, no CPUIF change), which preserves the canonical **blocking** read-stall exactly: the fast path accepts in cycle 0 (zero added latency), the slow path holds until the grant. Because the read blocks, SW physically cannot issue a second `next_id` read while one is pending, so the 1-deep latch is sufficient and no launch is ever dropped. Template-only, +30/-7.

Also fixes a latent multi-stream bug: `stream_idx_o` was keyed on the maskable req and reverted to 0 mid-stall; it is now keyed on `(seen | pending)`.

## Test

New self-checking `test/tb_idma_reg_frontend.sv` (APB DUT + controllable backend stub + `idma_transfer_id_gen` + a deadlock watchdog) with an `idma_sim_tb_idma_reg_frontend` make target — the first directed reg-frontend test in the repo. Cases: basic launch, backpressure/deadlock regression, multi-stream `stream_idx` hold, back-to-back monotonic ids.

- **With the fix:** PASS — 26 checks (NumStreams=1), 31 checks (NumStreams=2), 0 errors.
[...]

PR #154, opened 2026-07-09
frontend: Fix reg next_id launch deadlock (non-blocking)

DanielKellerM:
## What

Fixes the register-frontend `next_id` launch deadlock (introduced by the SystemRDL/PeakRDL migration in #73, inherited by #148) by making the launch **non-blocking**. Supersedes #153.

## Why non-blocking, not just deadlock-free

Reading `next_id` launches a transfer. With `next_id` modelled as a PeakRDL `external` register, that read *stalls the config bus* until the backend arbiter grants — which has two problems:

1. **Deadlock** — an ungranted read latches `external_pending`, masks the CPUIF request, and the read-ack can then never fire, so the whole reg block freezes.
2. **Head-of-line blocking** — even with the deadlock fixed, a *blocking* config read couples the config bus to backend readiness. On any SoC where the config port shares an interconnect with other traffic, a stalled `next_id` read back-pressures unrelated memory transactions — a real hang on integrated systems, not a corner case.

#153 fixed (1) but kept the blocking contract, so it still hits (2). A non-blocking read avoids both.

## The fix

`next_id` becomes an internal register with an `rd_swacc` read-strobe, so the config read **completes in the same cycle** (non-blocking — no HOL-block). A **launch-pending latch** then holds the arbiter request asserted until the backend grants, so a launch is **never dropped** when the grant lands a cycle after the (now-immediate) read. `status`/`done_id` are likewise internal, which also removes a latent freeze on reads of unused per-stream channel registers.

Template-only (`idma_reg.rdl` + `idma_reg.sv.tpl`); CPUIF-agnostic (apb/obi/axi4-lite).

## Software contract

The read no longer blocks — it returns immediately with the launch id. To serialise launches, software polls `next_id` until it advances (the launch was accepted) before programming the next transfer, instead of relying on the read to block.

## Test

New `test/tb_idma_reg_frontend.sv` (non-blocking contract) — the first directed reg-frontend testbench. It asserts the `next_id` read completes within a bounded latency **even under backpressure** (the non-blocking property), and that each returned id yields **exactly one** backend launch with no drop (the latch guarantee). Verified in Questa (NumStreams 1 and 2). Non-vacuity: the same testbench run against a *blocking* template fails the bounded-latency check (the read never completes under backpressure), so it genuinely distinguishes the contract.

## Supersedes / subsumes

- Supersedes #153 (blocking deadlock fix) — the blocking read-stall head-of-line-blocks on shared-interconnect SoCs.
- Subsumes #135 (internal `status`/`done_id`) — this rewrite makes them internal.


DanielKellerM:
Folded in the `stream_idx_o` arbitration bug @da-gazzi flagged (#150): `stream_idx_o` now rides the arbiter's winning index (`rr_arb_tree.idx_o`) instead of a last-wins scan, so with NumRegs>1 it always matches the arbitrated `dma_req_o`. Confirmed with a directed NumRegs=2 test (fails pre-fix, passes after). #150's `.ready` struct-vs-field fix is obsoleted here — this rewrite drops `proc_dma_backpressure`, so pready comes straight from the reg_top. The one remaining piece from #150 is the C-header / `target/sw` generation, which is orthogonal and can land separately. cc @da-gazzi — this now carries your fix; a review would be welcome.

src/frontend/reg/tpl/idma_reg.sv.tpl
```diff
@@ -98,17 +98,16 @@ module idma_${identifier} #(
...
-          for (int c = 0; c < NumStreams; c++) begin
-              if (dma_reg2hw[r].next_id[c].req && !dma_reg2hw[r].next_id[c].req_is_wr) begin
-                  stream_idx_o = c;
-              end
-          end
-      end
-  end
+  // per-port launch bookkeeping (see gen_core_regs: launch_pending latch)
+  logic    [NumRegs-1:0] launch_pending;
+  stream_t [NumRegs-1:0] held_stream;
```
da-gazzi: These guys don't follow the `name_q` naming convention for registers

src/frontend/reg/tpl/idma_reg.sv.tpl
```diff
@@ -98,17 +98,16 @@ module idma_${identifier} #(
...
+
+  // stream of the *arbitrated* port: must track dma_req_o (the arbiter winner), not the
+  // last launch_pending port, else the transfer is tracked under the wrong stream.
+  assign stream_idx_o = req_valid_o ? held_stream[arb_idx] : '0;
```
da-gazzi: I'm not sure about this one - if there is only one request and the arbiter accepts immediately, won't `req_valid_o` propagate to the output immediately, before the stream index is latched?

edit: I'm just seeing that the `valid` input handshake signal to the arbiter is latched with the same logic so that checks out, nevermind!

test/frontend/tb_idma_reg_frontend.sv
```diff
@@ -0,0 +1,728 @@
...
+  // Test 5 (multi-port arbitration) scoreboard state. Each config port programs a
+  // unique src_addr that encodes its target stream, so the arbiter's presented
+  // winner (dma_req_o) can be mapped back to a stream and compared to stream_idx_o.
+  logic [31:0] sb_addr_stream0;          // src_addr programmed for stream 0's port
+  logic [31:0] sb_addr_stream1;          // src_addr programmed for stream 1's port
+  int unsigned sb_mismatch;              // times stream_idx != the winner's stream
+  initial sb_mismatch = 0;
+
+  // --------------------------------------------------------------------------
+  // APB driver
```
da-gazzi: consider using the test infrastructure from the APB repo:
https://github.com/pulp-platform/apb/blob/master/src/apb_test.sv

da-gazzi:
Mostly LGTM, you could avoid some duplicate TB logic by using test infrastructure from the APB repo

src/frontend/reg/tpl/idma_reg.sv.tpl
```diff
@@ -98,17 +98,16 @@ module idma_${identifier} #(
...
-              end
-          end
-      end
-  end
+  // per-port launch bookkeeping (see gen_core_regs: launch_pending latch)
+  logic    [NumRegs-1:0] launch_pending;
+  stream_t [NumRegs-1:0] held_stream;
+
+  // stream of the *arbitrated* port: must track dma_req_o (the arbiter winner), not the
+  // last launch_pending port, else the transfer is tracked under the wrong stream.
```
DanielKellerM: Should be single line comment.

src/frontend/reg/tpl/idma_reg.sv.tpl
```diff
@@ -175,108 +174,127 @@ module idma_${identifier} #(
...
     );
 
-    logic read_happens;
-    // launch-stall: hold the reg read-ack until the arbiter accepts the request
-    // (protocol-agnostic — driven into hwif rd_ack below, see gen_hw2reg_connections)
+    // A software read of `next_id` (the non-external rd_swacc strobe) launches a transfer.
+    // The strobe pulses for a single cycle, so it is captured into `launch_pending` and
+    // held until the arbiter accepts the request (arb_ready, which implies req_ready_i).
+    // This keeps arb_valid asserted and the payload stable across a backend stall, honoring
+    // rr_arb_tree LockIn=1 (no dropped launch).
```
DanielKellerM: verbose comment

src/frontend/reg/tpl/idma_reg.sv.tpl
```diff
@@ -175,108 +174,127 @@ module idma_${identifier} #(
...
+        nxt_dma_req.d_req[${snd}].reps = 'd1;
 % endfor
       end
 % endfor
 % endif
     end
 
-    // observational registers
+    // observational registers (status/next_id/done_id are internal hw=w now: drive .next,
+    // read-side launch happens via the next_id rd_swacc strobe above)
```
DanielKellerM: verbose comment

test/frontend/tb_idma_reg_frontend.sv
```diff
@@ -0,0 +1,728 @@
...
+  localparam int unsigned DeadlockCycles = 32'd2000;
+
+  // register map (idma_reg32_3d_addrmap_pkg): base + per-stream stride 0x4
+  localparam logic [31:0] REG_CONF       = 32'h0000_0000;
+  localparam logic [31:0] REG_STATUS0    = 32'h0000_0004;
+  localparam logic [31:0] REG_NEXT_ID0   = 32'h0000_0044;
+  localparam logic [31:0] REG_DONE_ID0   = 32'h0000_0084;
+  localparam logic [31:0] REG_DST_ADDR   = 32'h0000_00D0;
+  localparam logic [31:0] REG_SRC_ADDR   = 32'h0000_00D4;
+  localparam logic [31:0] REG_LENGTH     = 32'h0000_00D8;
```
DanielKellerM, line 60: Is there some way we could use c generated headers from sys RDL? DPI-C

Bender.yml
```diff
@@ -118,6 +118,7 @@ sources:
       # Level 1
       - test/frontend/tb_idma_desc64_top.sv
       - test/frontend/tb_idma_desc64_bench.sv
+      - test/tb_idma_reg_frontend.sv
```
DanielKellerM: the test should be in frontend/

idma.mk
```diff
@@ -386,6 +386,17 @@ idma_sim_tb_idma_nd_midend_b2b: $(IDMA_VSIM_DIR)/compile.tcl
 	cd $(IDMA_VSIM_DIR); $(VSIM) -c -do "source compile.tcl; quit"
 	cd $(IDMA_VSIM_DIR); $(VSIM) -c -t 1ps -voptargs=+acc tb_idma_nd_midend_b2b -do "run -all; quit"
 
+# Self-checking reg-frontend regression. Checks the non-blocking next_id launch
+# contract (bounded-latency read + no dropped launch) across NumStreams 1/2, and
+# the multi-port arbitration of stream_idx with NumRegs=2, NumStreams=2.
+# Run with the Questa SEPP wrapper.
```
DanielKellerM: Remove comments

idma.mk
```diff
@@ -386,6 +386,17 @@ idma_sim_tb_idma_nd_midend_b2b: $(IDMA_VSIM_DIR)/compile.tcl
...
+.PHONY: idma_sim_tb_idma_reg_frontend
+idma_sim_tb_idma_reg_frontend: $(IDMA_VSIM_DIR)/compile.tcl
+	cd $(IDMA_VSIM_DIR); $(VSIM) -c -do "source compile.tcl; quit"
+	cd $(IDMA_VSIM_DIR); $(VSIM) -c -t 1ps -voptargs=+acc -gNumStreams=1 tb_idma_reg_frontend -do "run -all; quit"
+	cd $(IDMA_VSIM_DIR); $(VSIM) -c -t 1ps -voptargs=+acc -gNumStreams=2 tb_idma_reg_frontend -do "run -all; quit"
+	cd $(IDMA_VSIM_DIR); $(VSIM) -c -t 1ps -voptargs=+acc -gNumStreams=2 -gNumRegs=2 tb_idma_reg_frontend -do "run -all; quit"
```
DanielKellerM, line 394: can we find a more generic way of running the TBs

DanielKellerM:
Addressed the review in two new commits (`b3fa0f64`, `ad6b854d`) — appended on top, no force-push, the existing three untouched.

**Fixed**
- **`_q` naming** (@da-gazzi): `launch_pending`/`held_stream`/`arb_dma_req` → `*_q`; combinational nets left as-is.
- **Verbose comments** (tpl:109/181/286, `idma.mk`:392): tightened to a single line or removed.
- **Test location**: `git mv` → `test/frontend/`, grouped with the desc64 frontend TBs in `Bender.yml`.
- tpl:110 (@da-gazzi) — agreed it checks out; `arb_valid` is latched the same way, nothing to change.

**Deferred (follow-ups)**
- **C-headers via DPI-C** (tb:51): depends on the SystemRDL C-header generation (#150), which isn't in the repo yet — a natural follow-up once that lands.
- **Generic TB runner** (`idma.mk`:398): no minimal change short of a broader make refactor; leaving for a dedicated pass.

DanielKellerM:
@da-gazzi — took your `apb_test.sv` suggestion after all (feb5a512): the TB now drives via `apb_test::apb_driver`

src/frontend/reg/tpl/idma_reg.sv.tpl
DanielKellerM: Done.

src/frontend/reg/tpl/idma_reg.sv.tpl
DanielKellerM: Checks out, thanks.

test/frontend/tb_idma_reg_frontend.sv
DanielKellerM: Done; implemented in feb5a512.

src/frontend/reg/tpl/idma_reg.sv.tpl
DanielKellerM: Done.

src/frontend/reg/tpl/idma_reg.sv.tpl
DanielKellerM: Done.

src/frontend/reg/tpl/idma_reg.sv.tpl
DanielKellerM: Done.

test/frontend/tb_idma_reg_frontend.sv
DanielKellerM: Deferred → #155.

Bender.yml
DanielKellerM: Done.

idma.mk
DanielKellerM: Done.

idma.mk
DanielKellerM: Deferred → #156.

PR #160, opened 2026-07-20
Regs/Transpose: Add compute parameter register to register frontend and fix transpose to/from misaligned addresses

da-gazzi:
# Merge order:
This PR is based on the PR making the compute engine instantiation an elaboration-time parameter instead of a generation-time one, see https://github.com/pulp-platform/iDMA/pull/159 - that PR should be merged before this one.

# Changes:

- Add CI input for additional register frontends to generate in the `deploy` job (`IDMA_ADD_FE_IDS`)
- Add a register to hold compute (i.e., transpose) configuration to the register file definition
- BUGFIX: Transpose was buggy when the destination buffer was misaligned to the data bus
  - Previously, an output was popped from the transposition unit on every accepted output beat. As a misaligned write is split into 2 beats, that led to some data being dropped and AW transfers being issued without a corresponding W transfer to follow them up
  - This has been fixed by connecting the transpose unit's `ready_i` to `w_dp_req_ready` instead of `write_beat_done`; that signal is asserted on the last transaction of a write transfer, which is the correct behavior.

### DIFF
--- a/src/frontend/reg/idma_reg.rdl
+++ b/src/frontend/reg/idma_reg.rdl
@@ -151,6 +151,31 @@ addrmap idma_reg #(
         } reps [31:0] = 0;
     };
 
+    reg compute_cfg {
+        name = "compute_cfg";
+        desc = "Per-transfer on-the-fly compute configuration. The fields are sampled when next_id is read to launch a transfer.";
+        default sw = rw;
+        default hw = r;
+        field {
+            desc = "Enable on-the-fly compute for the launched transfer.";
+        } compute_enable [0:0] = 0;
+        field {
+            desc = "Compute operation selector. Currently, 1 selects transpose.";
+        } compute_op [4:1] = 0;
+        field {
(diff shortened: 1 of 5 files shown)


src/frontend/reg/idma_reg.rdl
```diff
@@ -151,6 +151,31 @@ addrmap idma_reg #(
         } reps [31:0] = 0;
     };
 
+    reg compute_cfg {
```
DanielKellerM, line 154: `compute_cfg` lands in every reg variant, not just the 4D transpose ones. Intentional? If so worth a release note; otherwise gate behind `NumDims>=4`.

DanielKellerM (approved):

src/frontend/reg/idma_reg.rdl
DanielKellerM: To be checked in the future

PR #166, opened 2026-08-06
backend: Restore per-beat w_beat_done for compute retire

DanielKellerM:
## Summary

The compute engine retires its output beats on the write manager's ready. Since #163 that is `w_dp_req_ready`, which for an AXI burst pulses only on the **last** beat: correct for byte-count-preserving compute (transpose issues single-beat bursts), but it replays beat 0 across every beat of a multi-beat write burst.

This partially reverts #163: `w_beat_done = write_happening` (strobe-independent, per-beat) is re-exposed from `idma_axi_write` and routed to the compute engine per active write protocol. AXI uses the per-beat pulse; other write ports keep `w_dp_ready`, which is already per-beat (e.g. OBI). The pulse becomes load-bearing for the multi-beat bursts that size-changing compute (follow-up PR) produces.

No behavior change for single-beat transpose bursts.

## Validation

- Builds standalone on `devel`; `tb_idma_transpose_b2b` green at DataWidth 32/64 (Questa 2026.1) at this commit alone.
- Generator rendered for all backend IDs plus `r_axil_w_axi`, `r_axis_w_axi`, `rw_tilelink`, and multihead variants: exactly one `w_beat_done` declaration and drive per topology.

### DIFF
--- a/src/backend/idma_axi_write.sv
+++ b/src/backend/idma_axi_write.sv
@@ -75,7 +75,9 @@ module idma_axi_write #(
     /// Ready to buffer
     output strb_t buffer_out_ready_o,
     /// External write-strobe mask (ANDed into wstrb); tie to '1 when unused
-    input  strb_t mask_ext_i
+    input  strb_t mask_ext_i,
+    /// Pulses when a write beat is accepted on the bus (strobe-independent)
+    output logic  w_beat_done_o
 );
     // offsets needed for masks to empty buffer
     strb_t w_first_mask;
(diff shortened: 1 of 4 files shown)


DanielKellerM:
Withdrawing this PR — deeper verification showed the per-beat retire is neither sufficient nor necessary:

- Transpose write bursts are single-beat by the midend strip decomposition (`NE*eb == StrbWidth`), where per-beat and per-transaction retire coincide. For (unsupported) multi-beat transpose bursts, *neither* signal streams correctly: per-transaction replays the first beat, per-beat desyncs the engine handshake. #167 now asserts the single-beat contract explicitly (`ComputeTransposeSingleBeat`) instead of silently relying on it.
- The size-changing MX ops in #167 no longer use a scalar retire at all: they pop lane-exactly from `buffer_out_ready`, which the write ports already generate per beat. No new signal is needed.

#167 has been rebased directly onto `devel` with transpose retire byte-identical to devel.

PR #167, opened 2026-08-06
backend: Add MX quant/dequant on-the-fly compute (size-changing datapath)

DanielKellerM:
## Summary

Adds size-changing on-the-fly compute (write byte count != read byte count) on top of the transpose/otf_compute framework, with MX (OCP microscaling) quantization and dequantization as the first ops. Based directly on `devel` (an earlier revision stacked on #166; the per-beat retire signal turned out to be unnecessary — see the retire note below — so #166 was withdrawn).

**Mechanism.** Per-op source:dest byte ratios in `idma_pkg` (`compute_in_bytes`/`compute_out_bytes`) program the legalizer write length, so the write burst's beat count matches the beats the compute engine emits. The compute seam presents per-lane valid (pack-buffer occupancy) with full strobes; the write manager's own beat mask gates acceptance, and the engine retires lane-exactly against the byte lanes the write port actually consumed (`buffer_out_ready`), so a partial tail beat pops only its own bytes — no drain FSM, no transport-layer state machine, and back-to-back transfers with an identical compute config pipeline with zero inter-transfer bubbles. The legalizer asserts fence the contract: input-granule alignment, beat-aligned src/dst, per-beat-capable destination protocols (no TileLink), the FP16 width cap, and that the requested op is elaborated (`EnableCompute` / `ComputeOps`).

**Retire note (transpose unchanged).** Transpose keeps devel's per-write-transaction retire (`w_dp_req_ready`) and its existing contract that every transpose write burst is single-beat (guaranteed by the midend strip decomposition, now made explicit by the `ComputeTransposeSingleBeat` assert — multi-beat transpose bursts mis-stream under *any* single-signal retire and were already unsupported on devel; a whole-tile "compact" transpose is future work). The MX ops do not use a scalar retire at all: they pop lane-exactly from `buffer_out_ready`, which the write ports already produce per beat. An earlier revision (#166) re-exposed a per-beat `w_beat_done` pulse for this; it is not needed and was dropped.

**Ops** (selected via `compute_cfg.compute_op`, same programming model as transpose):

| Op | Direction | Ratio | Width |
|---|---|---|---|
| `COMPUTE_MXQUANT` | FP32 -> MXFP8 | 128B -> 33B per 32-elem block | up to 1024b (StrbWidth 128) |
| `COMPUTE_MXQUANT_FP16` | FP16 -> MXFP8 | 64B -> 33B | up to 512b (StrbWidth 64) |
| `COMPUTE_MXDEQUANT` | MXFP8 -> FP32 | 33B -> 128B | length must be 33k with k % StrbWidth == 0 |

MX block layout is inline `[1B E8M0 scale][32B E5M2]`. The quantizer is RNE with full subnormal, saturation and Inf/NaN handling, bit-exact with the viDMA ALCU reference except two deliberate corrections (mirrored in the DPI golden): the E8M0 scale saturates instead of wrapping for all-tiny blocks, and Inf/NaN lanes are excluded from the shared-scale scan so a block's finite lanes survive. The split normal/subnormal rounding bands are load-bearing; do not merge them.

## Validation

Byte-exact against a DPI-C golden (`test/idma_mxquant_dpi.c`, pure integer pipeline) on Questa 2026.1 at every legal StrbWidth (DataWidth 32/64/128/256/512/1024):

- `tb_idma_mxquant`: FP16 and FP32 sources, a genuine partial 33B tail beat, a 4K page-crossing write.
- `tb_idma_mxroundtrip`: quant -> dequant back-to-back through one backend, byte-exact at both stages (the exact E5M2 identity); FP16 leg at 512b, FP32 leg at 1024b (256 blocks, 32KB legs).
- `tb_idma_mxrand`: constrained-random campaign — 40 serialized transfers per run with a random op each (FP16/FP32 quant, dequant, plain unaligned copy), random block counts, writes biased across 4K boundaries, behind an AXI shim injecting 40% random per-channel stalls; canary bytes fence each destination.
- `tb_idma_mxneg`: every compute guard assert proven to fire (11 cases, incl. the transpose single-beat and dequant length-overflow fences). Note this PR also adds `+define+INC_ASSERT` to the sim flow — without it `ASSERT_NEVER` compiles to nothing, so the legalizer guard asserts (including pre-existing ones) were inert in simulation.
- `tb_idma_transpose_b2b` / `tb_idma_transpose_nd` unchanged and green (backward compatibility).
- `tb_idma_mxperf`: steady-state throughput vs a plain-copy baseline on ideal memory — the bottleneck channel (quant: R, dequant: W) sustains 100% bus utilization at every StrbWidth, both for single transfers and for pipelined back-to-back streams (aggregate window across 8 transfers including the boundaries); the other channel sits at the compression ratio (e.g. 33/128). Three microarchitectural points make this hold: pack-buffer space only gates block-completing input beats, the dequant pack buffer carries one expanded-block window of refill-during-drain headroom, and lane-exact retire keeps tails sound across transfer boundaries. The pipelined quant stream is also checked byte-exact under the random-stall shim in `tb_idma_mxrand`.

## Known scope / disclosures

- Compute is functionally validated only on `idma_backend_rw_axi`. Size-changing compute on non-AXI src/dst is fenced with `NOT IMPLEMENTED` asserts (`ComputeMxSrcProtocol` / `ComputeMxDstProtocol`) until validated; OBI is the natural first candidate (its write ready is already per-beat). Transpose keeps its pre-existing protocol scope.
- Back-to-back compute transfers with an **identical** config pipeline freely (verified at 100% utilization). A config change (op, FP16 flag, params) is serialized **in hardware**: the request is held at the legalizer input until the datapath drained, so software needs no ordering rule. Verified with a different-config transfer issued pipelined behind a quant stream (byte-exact); the sub-unit $fatal asserts remain as backstops.
- Compute requires beat-aligned src/dst (asserted; all guard asserts covered by firing negative tests in `tb_idma_mxneg`).
- `ComputeOps` defaults to `'1`: existing `EnableCompute` designs gain the MX engines on regeneration (area); set `ComputeOps` explicitly to opt out.
- Synthesized standalone in a 12 nm technology at 1 GHz, timing met: quant 39.5k/71.5k cells (512b/1024b), dequant 37.4k/52.8k. The second commit restructures the engines for synthesis (hoisted completion datapath, generate-gated FP16, circular dequant buffer) — bit-exact and throughput-neutral, re-verified by the full suite. No in-system PnR data yet.
- Validated at `BufferDepth=3`, `MemSysDepth=0`, 1D transfers only.

### DIFF
--- a/Bender.yml
+++ b/Bender.yml
@@ -34,6 +34,7 @@ sources:
     files:
       # Level 0
       - src/idma_pkg.sv
+      - src/idma_mxquant_pkg.sv
       # Level 1
       - src/backend/idma_axil_read.sv
       - src/backend/idma_axil_write.sv
@@ -44,6 +45,8 @@ sources:
       - src/backend/idma_channel_coupler.sv
       - src/backend/idma_dataflow_element.sv
       - src/backend/idma_otf_transpose.sv
+      - src/backend/idma_otf_mxquant.sv
+      - src/backend/idma_otf_mxdequant.sv
       - src/backend/idma_otf_compute.sv
       - src/backend/idma_error_handler.sv
       - src/backend/idma_init_read.sv
@@ -135,6 +138,11 @@ sources:
       - target/rtl/tb_idma_generated.sv
       - test/tb_idma_transpose_nd.sv
       - test/tb_idma_transpose_b2b.sv
+      - test/tb_idma_mxquant.sv
+      - test/tb_idma_mxroundtrip.sv
+      - test/tb_idma_mxrand.sv
+      - test/tb_idma_mxneg.sv
+      - test/tb_idma_mxperf.sv
 
   # Multi-head directed backend testbenches
   - target: multihead
--- a/idma.mk
+++ b/idma.mk
@@ -360,6 +360,7 @@ $(IDMA_VSIM_DIR)/compile_%.tcl: $(IDMA_BENDER_FILES) $(IDMA_FULL_TB) $(IDMA_FULL
 IDMA_VLOG_ARGS  := -suppress vlog-2583 \
 			  	   -suppress vlog-13314 \
 			  	   -suppress vlog-13233 \
+			  	   +define+INC_ASSERT \
 			  	   -timescale \"1 ns / 1 ps\"
 
 define idma_generate_vsim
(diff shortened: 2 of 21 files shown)


DanielKellerM:
The merge was reverted off `devel` to fold in the wide-bus test coverage before landing. Superseded by #170, which re-adds this feature plus the audit fixes (block-scale clamp + Inf/NaN, dequant length fence, config-change interlock, wide-bus DW=256/512 coverage).

PR #168, opened 2026-08-07
spyglass: Fix comb-loop in `snitch`

gbellocchi:
This PR fixes a combinational loop found via Spyglass into the Snitch cluster. To accomplish this, I deactivated the `SameCycleRW` feature of `cc_passthrough_stream_fifo`, which is instantiated as a buffer in `idma_dataflow_element`.


gbellocchi:
This was fixed following a different approach, less invasive for the `idma`.

PR #171, opened 2026-08-11
backend: Add FP16-destination MX dequantization

DanielKellerM:
## Summary

Follow-up to #167: adds `COMPUTE_MXDEQUANT_FP16` (MXFP8 -> FP16, 33B -> 64B per 32-elem block) so the 512-bit configuration roundtrips FP16 -> MXFP8 -> FP16 natively, halving the expanded footprint vs the FP32 destination. Same programming model as the other ops (`compute_cfg.compute_op = 5`).

- The FP32 -> FP16 narrowing is IEEE (RNE, overflow saturates to Inf, quiet NaNs keep a payload bit). The C golden was cross-validated against numpy's `float32 -> float16` cast over 4.2M vectors (every exponent band, both signs, plus 4M randoms): zero mismatches.
- The dequant engine reuses the proven MXFP8 -> FP32 expansion and the circular pack buffer with a parametric insert length (64B/128B); no new datapath structure.
- New elaboration parameter `ComputeOps.mxfp16` gates the FP16 format paths (quant source *and* dequant destination) in or out per instance.
- Fenced: FP16 dequant above 512b (`ComputeMxdequantFp16Width`, one block per beat), covered by firing negative case 12; the beat-alignment and length-overflow fences extend to the new op.

## Validation

Questa 2026.1, byte-exact vs the DPI golden at every legal StrbWidth: roundtrip FP16 leg now returns FP16 (`tb_idma_mxroundtrip`), FP16-destination transfers mixed into the constrained-random stall campaign (`tb_idma_mxrand`), throughput gate green (`tb_idma_mxperf`: W = 100% at the destination bus, R at the 33:64 compression ratio), all 12 negative guards fire, transpose suites unchanged.

### DIFF
--- a/Bender.yml
+++ b/Bender.yml
@@ -34,6 +34,7 @@ sources:
     files:
       # Level 0
       - src/idma_pkg.sv
+      - src/idma_mxquant_pkg.sv
       # Level 1
       - src/backend/idma_axil_read.sv
       - src/backend/idma_axil_write.sv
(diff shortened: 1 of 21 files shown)


DanielKellerM:
Superseded by branch otf/mx-compute (78be850), which carries the FP16-destination dequant together with the full MX feature on current devel.

PR #170, opened 2026-08-11
backend: Add MX quant/dequant on-the-fly compute

DanielKellerM:
## Summary

Adds size-changing on-the-fly compute (write byte count != read byte count) to the otf framework, with MX (OCP microscaling) quantization and dequantization as the first ops. One commit on `devel`.

**Ops** (selected via `compute_cfg.compute_op`, same programming model as transpose):

| Op | Direction | Ratio (per 32-elem block) | Width |
|---|---|---|---|
| `COMPUTE_MXQUANT` | FP32 -> MXFP8 | 128B -> 33B | up to 1024b |
| `COMPUTE_MXQUANT_FP16` | FP16 -> MXFP8 | 64B -> 33B | up to 512b |
| `COMPUTE_MXDEQUANT` | MXFP8 -> FP32 | 33B -> 128B | up to 1024b |
| `COMPUTE_MXDEQUANT_FP16` | MXFP8 -> FP16 | 33B -> 64B | up to 512b |

MX block layout is inline `[1B E8M0 scale][32B E5M2]`; the 512b configuration roundtrips FP16 -> MXFP8 -> FP16 natively. Block geometry is single-homed in `idma_pkg` and the per-op byte ratios derive from it.

**Mechanism.** The per-op ratios program the legalizer write length, so the write burst's beat count matches what the engine emits. The compute seam presents per-lane pack occupancy with full strobes; the write manager's own beat mask gates acceptance, and the engine retires lane-exactly against the byte lanes the write port consumed (`buffer_out_ready`) — no drain FSM, no transport-layer state machine, and back-to-back transfers with an identical compute config pipeline with **zero inter-transfer bubbles at 100% bus utilization**. A config change is serialized in hardware at the request input (the request is held until the datapath drained, so its reads can never race a draining engine); software needs no ordering rules. Transpose keeps its per-write-transaction retire and single-beat burst contract, now asserted.

**Parameters.** `ComputeOps` (per-op elaboration mask; `mxfp16` gates the FP16 source/destination format paths — area opt-out) and `ComputeTuning` (engine tuning knobs). Element formats are typed (`mx_fmt_e`), derived from the opcode via a single-source function, like op support and byte ratios.

[...]

PR #173, opened 2026-08-11
backend: Add MX quant/dequant on-the-fly compute

DanielKellerM:
## Summary

Adds size-changing on-the-fly compute (write byte count != read byte count) to the otf framework, with MX (OCP microscaling) quantization and dequantization as the first ops. One commit on `devel`.

**Ops** (selected via `compute_cfg.compute_op`, same programming model as transpose):

| Op | Direction | Ratio (per 32-elem block) | Width |
|---|---|---|---|
| `COMPUTE_MXQUANT` | FP32 -> MXFP8 | 128B -> 33B | up to 1024b |
| `COMPUTE_MXQUANT_FP16` | FP16 -> MXFP8 | 64B -> 33B | up to 512b |
| `COMPUTE_MXDEQUANT` | MXFP8 -> FP32 | 33B -> 128B | up to 1024b |
| `COMPUTE_MXDEQUANT_FP16` | MXFP8 -> FP16 | 33B -> 64B | up to 512b |

MX block layout is inline `[1B E8M0 scale][32B E5M2]`; the 512b configuration roundtrips FP16 -> MXFP8 -> FP16 natively. Block geometry is single-homed in `idma_pkg` and the per-op byte ratios derive from it.

**Mechanism.** The per-op ratios program the legalizer write length, so the write burst's beat count matches what the engine emits. The compute seam presents per-lane pack occupancy with full strobes; the write manager's own beat mask gates acceptance, and the engine retires lane-exactly against the byte lanes the write port consumed (`buffer_out_ready`) — no drain FSM, no transport-layer state machine, and back-to-back transfers with an identical compute config pipeline with **zero inter-transfer bubbles at 100% bus utilization**. A config change is serialized in hardware at the request input (the request is held until the datapath drained, so its reads can never race a draining engine); software needs no ordering rules. Transpose keeps its per-write-transaction retire and single-beat burst contract, now asserted.

**Parameters.** `ComputeOps` (per-op elaboration mask; `mxfp16` gates the FP16 source/destination format paths — area opt-out) and `ComputeTuning` (engine tuning knobs). Element formats are typed (`mx_fmt_e`), derived from the opcode via a single-source function, like op support and byte ratios.

**Numerics.** The E8M0 block scale saturates at [-128,127] and excludes Inf/NaN lanes from its scan, so extreme blocks don't flush and one non-finite lane doesn't poison its block's finite values. The E5M2 quantizer is RNE with full subnormal, saturation and Inf/NaN handling. The FP32 -> FP16 narrowing is IEEE (RNE, overflow to Inf), cross-validated against numpy's float16 cast over 4.2M vectors: zero mismatches.

## Validation

Questa 2026.1, byte-exact against pure-integer DPI-C goldens (`test/golden/`) at every legal StrbWidth (DataWidth 32-1024; 32-512 in the make targets). Testbenches in `test/backend/` with a shared skeleton in `test/include/`:

- `tb_idma_mxquant`: FP16/FP32 sources incl. an all-tiny block, an Inf/NaN-poisoned block, FP specials, a genuine partial 33B tail beat and a 4K-crossing write.
- `tb_idma_mxroundtrip`: quant -> dequant through one backend, byte-exact at both stages; FP16 legs at 512b, FP32 at 1024b.
- `tb_idma_mxrand`: constrained-random campaign (random ops/blocks/addresses, 4K-biased writes) behind an AXI shim injecting 40% per-channel stalls, canary-fenced destinations, plus a pipelined same-config stream and a pipelined config-change transfer exercising the interlock.
- `tb_idma_mxneg`: all 13 guard fences proven to fire (alignment, widths, protocols, length overflow, transpose single-beat, un-elaborated op fatal). This PR also adds `+define+INC_ASSERT` to the sim flow — without it `ASSERT_NEVER` compiles to nothing.
- `tb_idma_mxperf`: bottleneck-channel utilization vs a plain-copy baseline — 100% at every width, single transfers and pipelined b2b streams; the other channel sits at the compression ratio.
- `tb_idma_transpose_b2b`/`_nd` unchanged and green.

Synthesized standalone in a 12 nm technology at 1 GHz, timing met: quant 39.5k/71.5k cells (512b/1024b), dequant 37.4k/52.8k.

## Known scope

- Compute is functionally validated on `idma_backend_rw_axi`; size-changing compute on non-AXI src/dst is fenced (`ComputeMxSrcProtocol`/`ComputeMxDstProtocol`) until validated. The stated path to lift this: a compute-aware job format in the generated `tb_idma_backend` template, giving every topology MX coverage through the existing harness.
- Identical-config transfers pipeline freely; config changes drain first (hardware-enforced, verified byte-exact).
- 1D transfers; ND rows work per the per-row fences (dst strides must keep rows beat-aligned). Whole-tile "compact" transpose remains future work.
- No in-system PnR data yet.


DanielKellerM:
Duplicate of #170 (opened without checking existing PRs — #170 now carries the consolidated state).

PR #184, opened 2026-08-12
ci: Elaborate the synthesis tops and inst64 frontend in public CI

DanielKellerM:
Public CI is lint and codegen only: no job elaborates RTL, so a port, parameter or connectivity break passes every gate. The EDA sims that would catch it run on the GitLab mirror and are skipped for fork PRs. PR #80 removing coupler ports is exactly this class.

Adds an `elaborate` workflow running `make idma_lint_all`:
- `idma_lint_elab` elaborates all 12 synthesis tops (8 backend variants, desc64, nd/mp/rt midends) with verilator.
- `idma_lint_inst64` elaborates the snitch_cluster-gated inst64 frontend through a concrete-typed wrapper (`test/idma_inst64_lint.sv`), since `idma_inst64_top` has all-`logic` parameters and no public concrete instance. inst64 was previously covered by nothing but syntax analysis.

### Verified locally (verilator 5.020, matching ubuntu-24.04)
- clean devel: all 12 tops plus inst64 elaborate, exit 0.
- non-vacuity: renaming `w_req_valid_i` in `idma_channel_coupler.sv` (a #80-class break) fails the gate with `Can't find definition of variable` and exit 2.

### Why not `pulp-actions/slang`
Evaluated it first. It is advisory by construction and cannot gate: `run_slang.py` exits 0 regardless of compile errors ("so the CI step continues and reviewdog can annotate"), and although reviewdog runs with `-fail-level=error`, that step is marked `continue-on-error: true`. It is a good complement for inline PR annotations and can be added alongside, but it would not fail a #80-class break. Happy to add it as an annotation layer if preferred.

### Side finding
Building the inst64 wrapper surfaced that `idma_inst64_events.sv` requires `obi_wr_req`/`obi_rd_req` members in the integrator-supplied `dma_events_t`, and that `ComputeEnable`/`testmode_i` no longer exist. There is no canonical public definition of `dma_events_t`, so snitch_cluster, the nonfree formal wrapper and this lint wrapper each hand-roll it and drift silently.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,6 +37,11 @@ jobs:
     uses: ./.github/workflows/analyze.yml
     secrets: inherit
 
+  elaborate:
+    needs: lint
+    uses: ./.github/workflows/elaborate.yml
+    secrets: inherit
+
   gitlab-ci:
     needs: build
     uses: ./.github/workflows/gitlab-ci.yml
--- a/.github/workflows/elaborate.yml
+++ b/.github/workflows/elaborate.yml
@@ -1,0 +1,47 @@
+# Copyright 2026 ETH Zurich and University of Bologna.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+# Author:
+# - Daniel Keller <dankeller@iis.ee.ethz.ch>
+
+name: elaborate
+
+on:
+  workflow_call:
+  workflow_dispatch:
+
+jobs:
+
+  # Elaborate every synthesis top plus the snitch_cluster-gated inst64 frontend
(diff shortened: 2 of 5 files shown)


DanielKellerM:
Superseded by #186, which carries this work and goes further: it folds the separate elaborate workflow into the verify matrix (they ran the same targets), installs Verilator through the shared setup action instead of a raw apt line, and replaces the lint-only inst64 wrapper with elaboration of the real testbench from #185.

Closing rather than merging: everything durable here (`idma_lint_elab`, `idma_lint_all`, `IDMA_LINT_TOPS`) is already in #186 and green there, while the two things this PR adds on top - `elaborate.yml` and `test/idma_inst64_lint.sv` - are deleted again by #186. Merging would be churn, and this is also the only one of the set with red checks.

PR #185, opened 2026-08-13
test: Add an inst64 frontend testbench driven over the accelerator bus

DanielKellerM:
The tightly-coupled inst64 frontend had no public testbench. Nothing exercised the DMA ISA decode, the transfer id and status path, or the datapath below it; only syntax analysis reached those sources.

Ports the existing harness (`idma_inst64_tb_pkg`, `idma_inst64_drv_if`, `idma_inst64_base`) and adds a directed AXI-to-AXI copy test driving dmsrc/dmdst/dmcpyi and polling dmstati.

### Verified
- Questa 2023.4: compiles 0 errors, reaches `$finish`, `TEST PASSED: 4096 B copied, ar=1 aw=1 beats`.
- Non-vacuity: corrupting one destination byte after the wait gives `payload mismatch at offset 777: expected 0xa9, got 0xba` and `TEST FAILED`. The TB also sentinels the destination and adds guard bands, so an unwritten byte or an overrun fails too.
- `make idma_sim_tb_idma_inst64_axi_copy` gates on both a negative `Error:|Fatal:` grep and a positive `TEST PASSED` grep, because Questa does not propagate `$fatal` to the exit code and a sim that never starts would pass a negative-only gate.

### Note on the port
The upstream harness builds on `DMOPC`, which is not part of the public ISA. It is not merely absent: `DMOPC` decodes bit-identically to the public `DMINIT`, so carrying it over would have launched a real zero-fill transfer, burning a transfer id and bumping `completed_id`, desynchronising the completion wait so it could return before the copy retired. The opcode and its callers are therefore dropped rather than stubbed. Head arguments are dropped for the same reason: the public decoder ignores `data_argb[12:5]`, so passing them would be a silent no-op.

### DIFF
--- a/idma.mk
+++ b/idma.mk
@@ -424,6 +424,18 @@ idma_sim_tb_idma_reg_frontend: $(IDMA_VSIM_DIR)/compile.tcl
 	cd $(IDMA_VSIM_DIR); $(VSIM) -c -t 1ps -voptargs=+acc -gNumStreams=2 tb_idma_reg_frontend -do "run -all; quit"
 	cd $(IDMA_VSIM_DIR); $(VSIM) -c -t 1ps -voptargs=+acc -gNumStreams=2 -gNumRegs=2 tb_idma_reg_frontend -do "run -all; quit"
 
+# Tightly-coupled inst64 frontend driven over the snitch accelerator bus.
+# Run with the Questa SEPP wrapper:
+#   make idma_sim_tb_idma_inst64_axi_copy VSIM="questa-2023.4 vsim"
+.PHONY: idma_sim_tb_idma_inst64_axi_copy
+idma_sim_tb_idma_inst64_axi_copy: $(IDMA_VSIM_DIR)/compile_tb_idma_inst64_axi_copy.tcl
+	cd $(IDMA_VSIM_DIR); $(VSIM) -c -do "source compile_tb_idma_inst64_axi_copy.tcl; quit"
+	cd $(IDMA_VSIM_DIR); $(VSIM) -c -t 1ps -voptargs=+acc tb_idma_inst64_axi_copy \
+		-logfile inst64_axi_copy.log -do "run -all; quit"
+	# Questa does not propagate $$fatal to the exit code; gate on the transcript
+	cd $(IDMA_VSIM_DIR); ! grep -qE "Error:|Fatal:" inst64_axi_copy.log
+	cd $(IDMA_VSIM_DIR); grep -q "TEST PASSED" inst64_axi_copy.log
+
 .PHONY: idma_sim_tb_idma_transpose_b2b
 idma_sim_tb_idma_transpose_b2b: $(IDMA_VSIM_DIR)/compile.tcl
 	cd $(IDMA_VSIM_DIR); $(VSIM) -c -do "source compile.tcl; quit"
--- a/idma.mk
+++ b/idma.mk
@@ -424,9 +424,6 @@ idma_sim_tb_idma_reg_frontend: $(IDMA_VSIM_DIR)/compile.tcl
 	cd $(IDMA_VSIM_DIR); $(VSIM) -c -t 1ps -voptargs=+acc -gNumStreams=2 tb_idma_reg_frontend -do "run -all; quit"
 	cd $(IDMA_VSIM_DIR); $(VSIM) -c -t 1ps -voptargs=+acc -gNumStreams=2 -gNumRegs=2 tb_idma_reg_frontend -do "run -all; quit"
 
-# Tightly-coupled inst64 frontend driven over the snitch accelerator bus.
-# Run with the Questa SEPP wrapper:
-#   make idma_sim_tb_idma_inst64_axi_copy VSIM="questa-2023.4 vsim"
 .PHONY: idma_sim_tb_idma_inst64_axi_copy
 idma_sim_tb_idma_inst64_axi_copy: $(IDMA_VSIM_DIR)/compile_tb_idma_inst64_axi_copy.tcl
 	cd $(IDMA_VSIM_DIR); $(VSIM) -c -do "source compile_tb_idma_inst64_axi_copy.tcl; quit"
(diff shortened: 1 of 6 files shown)


test/frontend/idma_inst64_drv_if.sv
```diff
@@ -0,0 +1,260 @@
...
+// Solderpad Hardware License, Version 0.51, see LICENSE for details.
+// SPDX-License-Identifier: SHL-0.51
+
+// Authors:
+// - Daniel Keller <dankeller@iis.ee.ethz.ch>
+
+/// Snitch accelerator-bus driver for `idma_inst64_top`. Exposes the public inst64 ISA
+/// only (DMSRC/DMDST/DMSTR/DMREP/DMCPY/DMCPYI/DMSTAT). The vidma-only DMOPC/ALU/MX
+/// opcodes are deliberately absent: DMOPC's encoding aliases public DMINIT and would
+/// silently launch a real zero-fill transfer.
```
DanielKellerM: Remove references to vidma The vidma-only DMOPC/ALU/MX
/// opcodes are deliberately absent: DMOPC's encoding aliases public DMINIT and would
/// silently launch a real zero-fill transfer.

test/frontend/idma_inst64_tb_pkg.sv
```diff
@@ -0,0 +1,133 @@
...
+// SPDX-License-Identifier: SHL-0.51
+
+// Authors:
+// - Daniel Keller <dankeller@iis.ee.ethz.ch>
+
+`include "axi/typedef.svh"
+`include "obi/typedef.svh"
+
+/// Concrete type binding for the `idma_inst64_top` testbenches. The types mirror
+/// `test/idma_inst64_lint.sv`, the elaboration-proven binding of the frontend.
```
DanielKellerM: remove this comments as they might be obsolete

test/frontend/tb_idma_inst64_axi_copy.sv
```diff
@@ -0,0 +1,336 @@
...
+    // sides to 0, so an equality proves nothing; needs a TCDM-window TB with an OBI
+    // read/write mix) and the per-channel index in gen_events (NumChannels == 1).
+    //
+    // Polarity note: aw_stall/ar_stall are valid && !ready, but idma_inst64_events
+    // overrides its own w_stall/r_stall at lines 95/96, so w_stall is a W-side bubble
+    // (w_ready && !w_valid) and r_stall is !r_ready && r_valid; the earlier assignments
+    // are dead code. Measured gap: none of the four stall fields ever asserts here, since
+    // the sim memories neither backpressure nor bubble. Their equality still catches a
+    // polarity flip or a spurious assert, but a stuck-at-0 stall would pass; that needs a
+    // ready-throttling harness, not a weaker check.
```
DanielKellerM: verbose explanation, reduce to 3 o4 4 lines

test/frontend/tb_idma_inst64_axi_copy.sv
```diff
@@ -0,0 +1,336 @@
...
+    ) else $fatal(1, "OBI leg requested during an AXI-to-AXI transfer: bad protocol decode");
+
+    task automatic seed_memories();
+        for (int i = 0; i < CopyPadBytes; i++) begin
+            harness.mem_write_byte(SrcAddr + i, (i < CopySize) ? (PatternStart + i) : 8'h00);
+        end
+        // Sentinel over the destination and its guard bands: an unwritten byte must fail
+        // the compare rather than accidentally match, and an overrun must be visible.
+        // Offsets stay non-negative; a signed offset added to a 64-bit unsigned address is
+        // zero-extended, not sign-extended.
```
DanielKellerM: verbose

idma.mk
```diff
@@ -424,6 +424,18 @@ idma_sim_tb_idma_reg_frontend: $(IDMA_VSIM_DIR)/compile.tcl
 	cd $(IDMA_VSIM_DIR); $(VSIM) -c -t 1ps -voptargs=+acc -gNumStreams=2 tb_idma_reg_frontend -do "run -all; quit"
 	cd $(IDMA_VSIM_DIR); $(VSIM) -c -t 1ps -voptargs=+acc -gNumStreams=2 -gNumRegs=2 tb_idma_reg_frontend -do "run -all; quit"
 
+# Tightly-coupled inst64 frontend driven over the snitch accelerator bus.
+# Run with the Questa SEPP wrapper:
+#   make idma_sim_tb_idma_inst64_axi_copy VSIM="questa-2023.4 vsim"
```
DanielKellerM: remove comments

DanielKellerM (commented):
fix

DanielKellerM:
Addressed in a1612d21.

- `idma_inst64_drv_if.sv`: dropped the vidma/DMOPC references; the docstring now just says what the driver covers.
- `idma_inst64_tb_pkg.sv`: removed the header comment. It pointed at `test/idma_inst64_lint.sv`, which #186 deletes, so it was already going stale.
- `tb_idma_inst64_axi_copy.sv:62`: 18 comment lines down to 5.
- `tb_idma_inst64_axi_copy.sv:194`: 4 lines down to 2.
- `idma.mk`: removed the target header block; no other sim target in the file carries one.

Re-verified after the edits: `make idma_sim_tb_idma_inst64_axi_copy` still passes, with the cross-check reporting the same 26 fields and totals.

One thing the edits surfaced, separate from this PR: `lint-sv` runs with `paths: src`, and the tree-scoped verible job in #186 also only walks `src/`, so **nothing lints `test/`**. Running verible over this testbench by hand reports `always-ff-non-blocking` on the counter updates, including ones that predate the cross-check. I have not touched them here since the blocking assignments are deliberate for the same-tick sampling, and changing them would need re-verification. Happy to extend the tree-scoped job to `test/` and clean these up in a follow-up if you want that coverage.

DanielKellerM:
Note on merge order: this now expects the documented stall polarity in the event cross-check, so **#187 has to land first**. On today's devel RTL the goldens for `r_stall`/`w_stall` would mismatch.

Background: an audit of #187 against `snitch_cluster_peripheral_reg.rdl` showed the RTL had the two stall fields overwritten by the buffer metrics, so both the RTL and this testbench's goldens were encoding the wrong polarity. Since the goldens mirrored the RTL, all four stall comparisons were `0 === 0` and could not fail. With #187 in place the corrected `r_stall` asserts 44 of 108 cycles and the comparison discriminates - proven by flipping the golden and watching it fail.

PR #186, opened 2026-08-13
ci: Verify the backend variants publicly with verilator and slang

DanielKellerM:
Public CI could not see a port, parameter or datapath defect. The real variant matrix runs on the private GitLab mirror with proprietary tools and is skipped for fork PRs, so external collaborators had no visibility and every check spent a licence. This adds a license-free public matrix that gates first.

### The matrix

- **Elaborate** all 8 backend variants under **both verilator and slang**, across their jobs.json parameter sets plus a width sweep and compute-enabled configurations.
- **Register frontend**: all three variants (`idma_reg32_3d`, `idma_reg64_2d`, `idma_reg64_1d`). Only the first was covered before.
- **inst64**: elaborated through the real testbench rather than a lint-only wrapper, which is deleted.
- **Simulate** the DPI-backed datapath tests byte-exact, and the negative guards with `--assert`.
- **Codegen and job-table consistency** checks.

The run set lives in `src/db/verify.yml`. `util/run_verify.py` runs a suite from it and emits the CI matrix from the same entries, and the workflow fans out with `fromJSON`, so a suite added to the database gets a CI leg with no workflow edit and nothing generated is committed. Testbench tops come from bender rather than a hand-kept list.

### The toolchain decides whether this works at all

`verilator --timing` lowers to C++20 coroutines and g++ 11 miscompiles them: the wide compute simulations SIGSEGV before time 0 with an empty log. Nothing recorded that, and the only green run worked because `CXX=g++-13.2.0` happened to be on the command line.

`idma.mk` now picks the compiler rather than inheriting it, folds it into `IDMA_VLT_MAKEFLAGS` instead of being replaced by a caller passing `-j4`, and `idma_verify_toolchain` fails with the reason instead of leaving a segfault to interpret.

This also retired two exclusions that were wrong. `mxquant` at DataWidth 512 and 1024 was listed as a verilator crash; it was the compiler, and both widths run now. Everything excluded is an `exclude:` entry in the database with a reason, and `check_jobs.py` fails when a negative case or a legalizer guard is neither exercised nor excluded.

### Breaking changes

Two, both downstream-visible, both in CHANGELOG.md:

- The tracer is one header per backend id. `idma/tracer.svh` keeps only the shared helpers; `IDMA_TRACER_<ID>` lives in `idma/tracer_<id>.svh`. Downstream replaces the include with the per-id header. This is what removed the recorded id-list stamp: the tracer covered every id in one file, so its content depended on the id list rather than on its own inputs.
- Out-of-tree `IDMA_ADD_IDS` variants no longer land in `target/rtl/idma_generated.sv`. They build via `idma_add_all` into `idma_generated_add.sv` and reach Bender through a new `add_ids` target. Overriding `IDMA_BACKEND_IDS` from the command line now needs a preceding `idma_rtl_clean`.

### On the width diagnostic in the legalizer

An earlier version of this description claimed `ComputeMxdequantLengthFits` carried a real RTL bug. It does not, and the claim was wrong.

`(64'd1 << $bits(req_i.length))` does overflow when the length field is 64 bits wide, and slang reports it where verilator does not. But the assertion conjoins `($bits(req_i.length) < 64)` ahead of it, so in the only case where the shift misbehaves the expression is already zero. The change to `65'd1` fixes a static width diagnostic and makes the intent explicit; it does not change behaviour, and the guard was never wrong at runtime.

### Verified

`make idma_verify_all` from a clean tree: rc=0, 24 elaboration OK, 23 sim PASS, 0 failures. Non-vacuity proven per gate by injection rather than assumed.

### Known limits, not closed here

- `tb_idma_mxroundtrip` above DataWidth 256 dies in a heap abort that valgrind traces into verilator's generated model constructor, before time 0, where `tb_idma_mxquant` is clean at the same width. Filed as #196; it also blocks a verilator 5.046 bump.
- `mxneg` cases 4 and 12 need DataWidth 1024, which SIGSEGVs on verilator 5.020 and passes on 5.046, so `ComputeMxFp16Width` has no firing test on the pinned toolchain.
- `mxneg` case 11 is never accepted, so `ComputeMxdequantLengthFits` is never sampled. `ComputeDstTilelink` has no TileLink variant to test against.
- `util/slang_elab.py` duplicates `pulp-platform/pulp-actions/slang`, which cannot fail a job today. Fix proposed upstream in pulp-platform/pulp-actions#41; tracked here by #197.
- The verilator leg keeps `-Wno-fatal` and promotes only LATCH/MULTIDRIVEN/IMPLICIT, so it is weaker than the slang leg, which runs `-Werror`.
- `IDMA_REG_CPUIF` is exercised only as `apb4-flat`.
- `idma_reg64_2d` and `idma_reg64_1d` gain elaboration only, never simulation.
- The 88 directed job files still need the licensed run: verilator cannot converge the class-based driver's virtual interface.

### Scope

This is larger than "elaborate with slang, simulate with verilator", and a review of the branch made the case that the tracer split and the `IDMA_ADD_IDS` rework are separate downstream-facing changes riding along in a CI PR. Both are here because removing the id-list stamp required them. If you would rather they landed on their own against 0.7.0, say so and I will split them out.


.github/workflows/ci.yml
```diff
@@ -37,8 +37,16 @@ jobs:
     uses: ./.github/workflows/analyze.yml
     secrets: inherit
 
+  verify:
+    needs: lint
+    uses: ./.github/workflows/verify.yml
+    secrets: inherit
+
+  # The licensed EDA pipeline runs only once the license-free matrix is green,
+  # so it confirms rather than being the only signal, and no licences are spent
```
DanielKellerM: remove verbose comments, max 1 line comment if really needed

.github/workflows/lint.yml
```diff
@@ -32,6 +32,24 @@ jobs:
         exclude_paths: |
           target/sim/vsim/wave/tpl/*.do.tpl
 
+  # Tree-scoped counterpart to lint-sv: that job reports through reviewdog with
+  # -diff, so it only sees lines a PR touched and pre-existing violations never
+  # fail it. This one checks all of src/ so the backlog stays at zero.
```
DanielKellerM: remove verbose comments, max 1 line comment if really needed

.github/workflows/verify.yml
```diff
@@ -0,0 +1,282 @@
...
+# and slang 11 (pyslang) only, so fork PRs - which never reach the proprietary EDA
+# sims on the GitLab mirror - get a real gate, and the licensed run becomes a
+# confirmation rather than the only signal.
+#
+# Two tiers: elaboration first, simulation behind it. elab-backend, elab-shared-tops
+# and elab-tb-shared run in parallel and every simulation job `needs:` all three, so a
+# port, parameter or connectivity break fails in seconds instead of burning every
+# simulation runner. `fail-fast: false` stays on the matrices, so the elaboration tier
+# still reports every leg. elab-multihead, codegen-consistency and report-slang-extra
+# are parallel peers, not part of the barrier: they gate nothing the simulations use.
```
DanielKellerM: remove verbose comments, max 1 line comment if really needed

.github/workflows/verify.yml
```diff
@@ -0,0 +1,282 @@
...
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+# Author:
+# - Daniel Keller <dankeller@iis.ee.ethz.ch>
+
+# License-free verification matrix. It runs on Verilator 5.020 (apt, ubuntu-24.04)
+# and slang 11 (pyslang) only, so fork PRs - which never reach the proprietary EDA
+# sims on the GitLab mirror - get a real gate, and the licensed run becomes a
+# confirmation rather than the only signal.
```
DanielKellerM: remove verbose comments, max 1 line comment if really needed

.github/workflows/verify.yml
```diff
@@ -0,0 +1,282 @@
...
+name: verify
+
+on:
+  workflow_call:
+  workflow_dispatch:
+
+jobs:
+
+  # Codegen hygiene, not verification: jobs.json and the generated tree must
+  # still describe the same design, and a second generation must be identical.
```
DanielKellerM: remove verbose comments, max 1 line comment if really needed

.github/workflows/verify.yml
```diff
@@ -0,0 +1,282 @@
...
+      -
+        name: Set up toolchain
+        uses: ./.github/actions/verify-setup
+      -
+        name: Check jobs.json, the CI matrix and codegen determinism
+        run: uv run --locked make idma_verify_codegen
+
+  # Elaboration of one backend variant under both front ends, with the real
+  # per-variant jobs.json parameters plus a DataWidth sweep, and slang
+  # elaboration of that variant's testbench.
```
DanielKellerM: remove verbose comments, max 1 line comment if really needed

.github/workflows/verify.yml
```diff
@@ -0,0 +1,282 @@
...
+      -
+        name: Generate RTL
+        run: uv run --locked make idma_hw_all
+      -
+        name: Elaborate ${{ matrix.id }}
+        run: uv run --locked make idma_verify_backend IDMA_VERIFY_ID=${{ matrix.id }}
+
+  # The four non-backend synthesis tops with their own jobs.json parameters, plus
+  # the default-parameter elaboration of all 12 tops and of the inst64 testbench,
+  # the only public concrete binding of idma_inst64_top.
```
DanielKellerM: remove verbose comments, max 1 line comment if really needed

.github/workflows/verify.yml
```diff
@@ -0,0 +1,282 @@
...
+      -
+        name: Generate RTL
+        run: uv run --locked make idma_hw_all
+      -
+        name: Elaborate the shared synthesis tops
+        run: uv run --locked make idma_verify_shared
+
+  # Testbench tops that are not per-backend, including the three generated register
+  # frontends through tb_idma_reg_frontend. Verilator cannot parse the verification
+  # stack at all, so slang is the only public tool for this half of the repository.
```
DanielKellerM: remove verbose comments, max 1 line comment if really needed

.github/workflows/verify.yml
```diff
@@ -0,0 +1,282 @@
...
+        name: Set up toolchain
+        uses: ./.github/actions/verify-setup
+        with:
+          verilator: 'true'
+      -
+        name: Elaborate the multi-head variants and testbenches
+        run: uv run --locked make idma_verify_multihead
+
+  # Real simulation, byte-exact against the DPI-C goldens. The destination is
+  # pre-poisoned, so a backend that writes nothing fails.
```
DanielKellerM: remove verbose comments, max 1 line comment if really needed

.github/workflows/verify.yml
```diff
@@ -0,0 +1,282 @@
...
+        name: Set up toolchain
+        uses: ./.github/actions/verify-setup
+      -
+        name: Generate RTL
+        run: uv run --locked make idma_hw_all
+      -
+        name: Elaborate the shared testbenches
+        run: uv run --locked make idma_verify_tb_shared
+
+  # Out-of-tree multi-head build; license-free but generated only on request.
```
DanielKellerM: remove verbose comments, max 1 line comment if really needed

.github/workflows/verify.yml
```diff
@@ -0,0 +1,282 @@
...
+      -
+        name: Generate RTL
+        run: uv run --locked make idma_hw_all
+      -
+        name: Simulate the transpose datapath, plain and backpressured
+        run: uv run --locked make idma_verify_sim_transpose IDMA_VLT_MAKEFLAGS=-j4
+
+  # Negative tests, inverted polarity: the run must exit non-zero AND name its
+  # guard. --assert is load-bearing; without it both testbenches exit 0 and the
+  # leg could not fail.
```
DanielKellerM: remove verbose comments, max 1 line comment if really needed

src/backend/tpl/idma_legalizer.sv.tpl
```diff
@@ -726,13 +726,14 @@ ${database[protocol]['legalizer_write_data_path']}
                    (req_i.opt.compute.op == idma_pkg::COMPUTE_MXDEQUANT_FP16)) &
                   (req_i.length % (idma_pkg::MxBlockBytes*StrbWidth) != 0)), clk_i, !rst_ni)
     // NOT IMPLEMENTED: dequant output length that overflows the length field
+    // 65-bit literal: a 64-bit length shifts by 64, which overflows a 64-bit operand
```
DanielKellerM: remove verbose comments, max 1 line comment if really needed

test/frontend/tb_idma_reg_frontend.sv
```diff
@@ -6,6 +6,8 @@
 // - Daniel Keller <dankeller@iis.ee.ethz.ch>
 
 // Self-checking testbench for the iDMA register frontend (idma_reg32_3d, apb4-flat).
+// `RegVariant` also binds idma_reg64_2d and idma_reg64_1d for elaboration only; they
+// have no stimulus here, and are covered by nothing else.
```
DanielKellerM: remove verbose comments, max 1 line comment if really needed

test/frontend/tb_idma_reg_frontend.sv
```diff
@@ -20,7 +22,11 @@ module tb_idma_reg_frontend import idma_pkg::*; import apb_test::apb_driver; #(
   // number of streams the elaborated DUT exposes (checked at instantiation)
   parameter int unsigned NumStreams = 32'd1,
   // number of config-bus ports (arbitrated by the reg frontend's rr_arb_tree)
-  parameter int unsigned NumRegs    = 32'd1
+  parameter int unsigned NumRegs    = 32'd1,
+  // generated frontend under test: 0 = idma_reg32_3d, 1 = idma_reg64_2d, 2 = idma_reg64_1d.
+  // Only variant 0 is simulated; 1 and 2 are elaboration-only (their register maps differ
+  // and reg64_1d emits a flat 1D request), and the stimulus refuses to run on them.
```
DanielKellerM: v

test/frontend/tb_idma_reg_frontend.sv
```diff
@@ -33,11 +39,13 @@ module tb_idma_reg_frontend import idma_pkg::*; import apb_test::apb_driver; #(
...
   localparam int unsigned CfgStrbWidth   = CfgDataWidth / 32'd8;
   localparam int unsigned IdCounterWidth = 32'd32;
-  // idma data-path (reg32_3d: 32-bit data, 3 ND dims)
-  localparam int unsigned AddrWidth      = 32'd32;
-  localparam int unsigned DataWidth      = 32'd32;
-  localparam int unsigned NumDim         = 32'd3;
-  localparam int unsigned RepWidth       = 32'd32;
+  // idma data-path: reg32_3d is 32-bit over 3 ND dims, both reg64 variants are 64-bit.
+  // NumDim shapes the ND request type; reg64_1d takes the flat idma_req_t instead, so
+  // its ND type is unused and the 2-dim shape is kept only to keep the macro legal.
```
DanielKellerM: remove verbose comments, max 1 line comment if really needed

test/frontend/tb_idma_reg_frontend.sv
```diff
@@ -50,7 +58,8 @@ module tb_idma_reg_frontend import idma_pkg::*; import apb_test::apb_driver; #(
   // config-clock cycles is a hang (the non-blocking read completes immediately).
   localparam int unsigned DeadlockCycles = 32'd2000;
 
-  // register map (idma_reg32_3d_addrmap_pkg): base + per-stream stride 0x4
+  // register map (idma_reg32_3d_addrmap_pkg): base + per-stream stride 0x4; the reg64
+  // variants have their own map and are elaboration-only, so it is not modelled here
```
DanielKellerM: remove verbose comments, max 1 line comment if really needed

test/frontend/tb_idma_reg_frontend.sv
```diff
@@ -179,27 +188,86 @@ module tb_idma_reg_frontend import idma_pkg::*; import apb_test::apb_driver; #(
...
-    .next_id_i      ( next_id     ),
-    .stream_idx_o   ( stream_idx  ),
-    .done_id_i      ( done_id     ),
-    .busy_i         ( busy        ),
-    .midend_busy_i  ( midend_busy )
-  );
+  // All three generated frontends share the parameter and port list; only the module
+  // name and the emitted request type differ. reg64_1d (NumDim=1) emits a flat
+  // idma_req_t, which is widened back to the ND shape the stimulus works on.
+  if (RegVariant == 32'd0) begin : gen_reg32_3d
```
DanielKellerM: remove verbose comments, max 1 line comment if really needed

util/mario/tracer.py
```diff
@@ -120,19 +120,20 @@ def render_tracer(prot_ids: dict, db: dict, tpl_file: str) -> str:
         # signals
         signals = ''
 
-        # handle read ports
+        # Keys are direction-qualified: a protocol present on both sides (INIT) otherwise
+        # emits the same key twice and the write leg silently drops the read channel.
```
DanielKellerM: remove verbose comments, max 1 line comment if really needed

DanielKellerM (commented):
in general lgtm, but fix commnets

DanielKellerM:
Comments trimmed in `ecdaebc8`. Every comment this PR adds is now a single line:

| file | added comment lines |
|---|---|
| `.github/workflows/ci.yml` | 1 |
| `.github/workflows/lint.yml` | 2 |
| `.github/workflows/verify.yml` | 14 (was a 38-line header plus 7 multi-line blocks) |
| `util/mario/tracer.py` | 2 |
| `test/frontend/tb_idma_reg_frontend.sv` | 7 |
| `src/backend/tpl/idma_legalizer.sv.tpl` | 1 |

Two things worth flagging.

**The 38-line header carried a NAMED EXCLUSIONS list** - the six testbenches that are elaborated but not simulated, the DataWidth 512/1024 SIGSEGV cases, mxneg 4/11/12, and the 88 directed job files. That detail is now in the commit message and the PR body rather than the file. Say the word if you would rather it lived somewhere in-tree; my worry is that a reader of `verify.yml` can no longer see what is deliberately not covered, but a 38-line banner was not the right home for it either.

**I over-reached once and backed it out.** My first pass collapsed comment blocks across the whole of `idma_legalizer.sv.tpl` and `tb_idma_reg_frontend.sv`, including pre-existing comments this PR never touched - 17 and 36 blocks. That would have buried the real change in unrelated churn, so I reverted both files and re-did it against the diff, touching only comments this PR introduces. The legalizer's single added comment was already one line and is untouched.

Re-verified after trimming: `make idma_verify_all` rc=0, 24 elaboration OK, 21 sim PASS, 0 failures; yamllint clean on all three workflows; flake8 clean on `tracer.py`.

#192 is restacked on top. That rebase hit a conflict where this PR's comment trim met #192's `parameter-name-style` renames in `tb_idma_reg_frontend.sv`; resolved by keeping the renames and leaving the trimmed comment out, with no stale `REG_*` references remaining and `idma_verify_all` green on that branch too.

PR #193, opened 2026-08-13
backend: Hold the zero-length rejection until it is accepted

### LINKED ISSUE #189: Backend drops the zero-length rejection response when the consumer is not ready
## Summary
With `RejectZeroTransfers=1`, the backend answers a zero-length request combinationally in the cycle it accepts it, but acceptance does not depend on `rsp_ready_i`. The injected error response is asserted for exactly one cycle, so if the consumer is not ready in that cycle the response is **silently dropped**. The backend has then accepted N requests and delivered N-1 responses; every later response is credited to the wrong transfer, and the last one is never acked.

`gen_reject_zero_transfers` in `src/backend/tpl/idma_backend.sv.tpl`:
```systemverilog
if (is_length_zero & req_valid_i & req_ready_o) begin
    rsp_ready = 1'b0;
    rsp_valid_o = 1'b1;   // one cycle only, regardless of rsp_ready_i
    ...
end
```

## Affected
Every variant built with `RejectZeroTransfers=1`; the construct is variant-independent. Reproduced on `tb_idma_backend_rw_axi_rw_axis`, and the same stimulus also reproduces on `rw_axi` and `r_obi_w_axi`. With `RejectZeroTransfers=0` the identical stimulus passes.

## Minimal reproduction
Two transfers, no randomisation needed:

| | length | src | dst | src/dst proto | max_src_len | max_dst_len | decouple aw/rw |
|---|---|---|---|---|---|---|---|
| A | 16 | 0x1000 | 0x80001000 | 0 / 0 | 1 | 128 | 0 / 0 |
| B | **0** | 0x2000 | 0x80002000 | 0 / 0 | 256 | 256 | 0 / 0 |

Transfer A must split into enough bursts that the legalizer retires while the datapath is still draining, so that the consumer is busy when B's rejection response is emitted. Boundary behaviour: `length=8` with `max_src_len=1` passes, `length=16` fails; `max_src_len=2` passes; `max_src_len=256/max_dst_len=256` passes at any length.

Deterministic, trips in 1360 ns.

## Observed stuck state
`dp_busy=1`; `r_dp_req_out_valid=1` with ready low and `w_dp_req_out_valid=1` with ready low (both FIFOs at usage 1); `i_idma_axi_write.w_dp_valid_i=1`, `w_dp_ready_o=0`, `aw_valid_i=0` - the write transport waits for an AW meta that never comes; `r_dp_rsp_valid` and `w_dp_rsp_valid` stay high with ready low forever. The legalizer is idle and `req_ready_o=1`. On the memory port, `axi_rsp_mem.b_valid=1` is never taken and three AW bursts (= `NumAxInFlight`) never receive a B.

## Not fixed by the existing interlock
A `zero_len_stall` interlock addresses the other half of this defect, where the injected response overtakes an older in-flight transfer's response. It does **not** gate acceptance on `rsp_ready_i`, so this lost-response hole survives it.

## Possibly a second instance
`src/midend/idma_nd_midend.sv:99` and `212-228` show the same accept-without-checking-ready shape with neither interlock present. Flagged by inspection only; four directed ND shapes failed to provoke it, so it is unproven and needs its own investigation.

## Impact on CI
This is the one remaining red `vsim-sim-random` leg (`rw_axi_rw_axis` seed 1). The other 10 were two testbench defects, fixed in #188. After that PR the leg fails honestly on this bug rather than on a bogus mismatch.

DanielKellerM:
Fixes #189.

With `RejectZeroTransfers=1` the rejection response was a purely combinational function of the **request** handshake, so it carried no state. That gave it two failure modes:

- **Lost response.** Its assertion window is one cycle and is not tied to `rsp_ready_i`, so a consumer that is busy elsewhere never sees it. The backend has then accepted N requests and delivered N-1 responses, and every later response is credited to the wrong transfer.
- **Reordered response.** Its position in the response stream is not ordered against transfers still in flight, so it can overtake an older transfer's response.

## The fix
Latch the rejection into a proper valid/ready stream that is held until accepted, and hold back a zero-length request while an older rejection is unaccepted or any transfer still owes a response:
```systemverilog
assign zero_len_stall = is_length_zero & (zero_rsp_pending_q | (num_outst_q != '0));
assign req_ready_o    = leg_ready & ~zero_len_stall;
```

Gating acceptance directly on `rsp_ready_i` was considered and rejected: it puts a bus-facing ready into the `req_ready_o` cone and closes a real combinational loop through the ND midend and the inst64 top. Here `zero_len_stall` depends only on flops, so no loop is created. Confirmed with verilator: the UNOPTFLAT warnings on this top are pre-existing and unchanged in count.

## Verified
- **Directed regression added** (`jobs/backend_*/zero_transfer_busy.txt`, registered in `jobs.json`): deterministic, no randomisation. Passes on rw_axi, rw_axi_rw_axis, r_obi_w_axi, r_axi_w_obi and rw_obi.
- **Non-vacuity**: reverting only the RTL fix and keeping the test reproduces the hang - `Tripped Watchdog (i_axi_r_watchdog)` at 1170 ns on rw_axi and 1160 ns on rw_axi_rw_axis. Restoring it passes again. Note `r_axi_w_obi` and `rw_obi` pass either way, since an OBI write destination commits synchronously and the deadlock chain cannot close.
- **The CI symptom**: the `rw_axi_rw_axis` random seed 1 leg, which contains two zero-length transfers, goes from failing (`Tripped Watchdog` at 236210 ns, 19 of 20 jobs launched) to passing (20 of 20, normal finish).
- **Codegen invariant**: the template is the only RTL edit; a forced regeneration reproduces `target/rtl` byte-for-byte, so nothing generated was hand-edited.
- All verdicts are gated on the transcript rather than the exit code; every failing run above still returned vsim exit 0 while carrying `** Fatal:`.

## Follow-up, not fixed here
`src/midend/idma_nd_midend.sv` (around lines 211-227) has the same construct with neither interlock. It was flagged by inspection and four directed ND shapes failed to provoke it, so it is unproven and deserves its own investigation rather than a speculative copy of this fix.


DanielKellerM:
Deferred until after the 0.7.0 release, per Daniel. Not to be merged into the release branch.

DanielKellerM:
Moved to #247 so the internal CI runs on it.

PR #195, opened 2026-08-13
backend: Accept a zero-length transfer as a no-op when RejectZeroTransfers is 0

### LINKED ISSUE #194: Zero-length transfers are unhandled with RejectZeroTransfers=0
`RejectZeroTransfers` selects whether the backend rejects a zero-length request or accepts it, but only the rejecting setting is exercised or modelled. With `RejectZeroTransfers=0` a zero-length transfer does not behave correctly and nothing in CI covers it.

## Reproduction
Two transfers to an AXI-Stream destination, the first zero-length:

| | length | src | dst | src/dst proto | max_src_len | max_dst_len |
|---|---|---|---|---|---|---|
| A | **0** | 0x1000 | 0x80001000 | 0 / 5 | 256 | 256 |
| B | 64 | 0x2000 | 0x80002000 | 0 / 5 | 256 | 256 |

On `tb_idma_backend_rw_axi_rw_axis`:
```
-GRejectZeroTransfers=1  ->  PASS
-GRejectZeroTransfers=0  ->  FAIL, Mismatch!
```
This predates #188; I confirmed it on unmodified devel as well as with that PR applied, so it is neither caused nor worsened by it.

## What is unclear
It is not established what `=0` is supposed to mean end to end:
- Does the backend emit any bus beats for a zero-length transfer, or none?
- Does it emit a response, and with what status?
- Is the legaliser expected to produce a degenerate burst, or nothing at all?

Until that is settled it is not possible to say whether the observed mismatch is a testbench modelling gap, a legaliser issue, or an unsupported configuration that should be rejected at elaboration.

## Current state after #188
The AXI-Stream write model skips zero-length jobs, which is correct only for the rejecting configuration. #188 makes that assumption explicit: with `=0` it now fails with `zero-length stream transfer is not modelled with RejectZeroTransfers=0` instead of a misleading data mismatch. That is a guard, not a fix.

## Why it stayed hidden
The CI matrix only ever runs `RejectZeroTransfers=1`. Every variant in `jobs.json` uses the rejecting setting, so the alternative has no coverage at all.

## Related
#189 and #193 cover the response handling of a *rejected* zero-length transfer, i.e. the `=1` path. Whatever is decided here likely touches the same `gen_reject_zero_transfers` region and its non-rejecting counterpart.

DanielKellerM:
Fixes #194. Stacked on #193 (both restructure the same generate block); merge after it.

`RejectZeroTransfers=0` is not merely unmodelled, it **silently corrupts memory**. The zero-length request is passed straight to the legaliser, which emits one read and one write burst descriptor with `num_bytes = 0` - and zero has no representation downstream:

- `AxLEN = (num_bytes + offset - 1) >> OffsetWidth` **underflows** at offset 0, giving a 256-beat, 1024-byte burst
- `tailer = (num_bytes + offset)[OffsetWidth-1:0] = 0` aliases to "all bytes valid", so WSTRB is `1111` on every beat

Measured on `tb_idma_backend_rw_axi`, one `length=0` job at `0x1000 -> 0x2000`: AR len=255, AW len=255, **256 write beats with full strobes**, B OKAY, and an iDMA response with **error=0**. A zero-length transfer reads 1024 bytes of unrelated source memory, overwrites 1024 bytes at the destination, and reports success. The testbench passes, because `compare_mem` with length 0 checks nothing.

Other shapes are worse: on OBI it writes 4 bytes of **undefined data**; with a misaligned source it **hard deadlocks** (watchdog at 1020 ns, no response ever); and on AXI Stream the read and write beat counts diverge, so 1020 bytes of over-read leak into the **next** job, which is the only reason the bug was visible at all.

`is_length_zero` is also left undriven in the non-rejecting branch and reads X for the whole simulation.

## The fix
The invariant is the **response contract**, not the byte count: one accepted request owes exactly one response, and the ND/RT midends and the descriptor frontends count on that. So both settings must suppress the request at the input and synthesise the response; `RejectZeroTransfers` legitimately selects only the payload:

| | response |
|---|---|
| `=1` | error=1, err_type=BACKEND |
| `=0` | error=0, a no-op that succeeds |

It can never legitimately hand the request to the legaliser, because no supported protocol can express a zero-beat transaction and AXI has no side-effect-free zero-byte read. The change is one generate block hoisted out of the `if (RejectZeroTransfers)`; no legaliser, transport-layer, database or testbench-model change.

## Verified
- The directed zero-length job passes at **both** settings, and 16 zero-length runs (8 job and testbench combinations x 2 settings) pass.
- Exactly one response per zero-length request at both settings, with the payloads above, probed on `idma_rsp_o` at the handshake.
- No AR, AW or W handshake at either setting; the probe was self-checked against a real transfer so the absence is real.
- The two-job AXI-Stream case that reported `Mismatch!` on devel at `=0` now passes; I re-ran that one myself, before and after.
- No regression: the shipped directed suites still pass, and the change elaborates under verilator and slang.

## Coverage note
Every instantiation in the repo hardwires `RejectZeroTransfers = 1'b1` - jobs.json, all testbenches, `idma_inst64_top`, `idma_nd_midend_synth` and both template defaults - so `=0` had no in-repo user and no CI coverage. That is why this survived.


DanielKellerM:
Deferred until after the 0.7.0 release, per Daniel. Not to be merged into the release branch.

PR #233, opened 2026-09-21
frontend: Make the inst64 TCDM and OBI topology a parameter

DanielKellerM:
#88 replaced the plain `rw_axi` instantiation in `idma_inst64_top` with an OBI plus INIT backend, so the frontend now demands an OBI port and an address map. `snitch_cluster` main pins iDMA 0.6.5, is AXI-only, and therefore cannot adopt 0.7.0 at all; that is what ties the bump to snitch_cluster#330. Occurrences of `obi_req_o` in that file: 0 at v0.6.4, 0 at v0.6.5, 6 at v0.7.0.

`EnableTcdmObi` selects the backend again. The default is `1'b1`, so nothing existing changes.

```systemverilog
idma_inst64_top #(
  .EnableTcdmObi (1'b0),   // idma_backend_rw_axi, no OBI port, no address decode
  ...
```

Guarding only the ports is not enough: the OBI and INIT type parameters default to `logic`, and `idma_obi_write.sv:111` and `idma_inst64_events.sv` do member selects on them, which are type-checked regardless of reachability. Both sit behind generates. `DMINIT` has no INIT path in this topology, so it is rejected with an error response rather than silently degrading to an AXI read.

`idma.mk` unions the two ids inst64 selects into the generated tracer set; a trimmed `IDMA_BACKEND_IDS` previously dropped a header the frontend includes unconditionally, which already broke inst64 on `devel`.

Verified: `idma_lint_inst64` elaborates four configurations (default, `EnableTcdmObi=0`, `DMATracing=1`, TCDM); `tb_idma_inst64_axi_copy`, `idma_sim_tb_idma_inst64_axi_only` and the new `tb_idma_inst64_tcdm_copy` pass on Questa 2026.1, the last moving 32 OBI write and 32 OBI read beats. Breaking the OBI write mux or the INIT memset data fails only the TCDM test, so the AXI pair really is blind to that leg. `idma_verify_codegen`, `idma_lint_sv`, `lint-commits` clean. Default topology is byte-identical to the pre-change run.

nonfree needs a lockstep change for the new generate scope level, and the `idma_sim_tb_idma_inst64_*` targets are not in any CI job yet.

Relates to snitch_cluster#330. #231 rewrites the same `TCDMAliasEnable` region, so it should land first.

Issue #232, opened 2026-09-17
backend: Synth wrappers never drive opt.compute, so compute optimises away


DanielKellerM:
The generated `idma_backend_synth_*` wrappers expose `EnableCompute` / `ComputeOps` / `ComputeTuning` but never drive `idma_req.opt.compute`. There is no `req_compute_*_i` port and no assign.

`src/backend/tpl/idma_backend_synth.sv.tpl:380-411` assigns every `opt.*` field except `compute`. In the v0.7.0 deploy, `target/rtl/idma_backend_synth_rw_axi.sv:296-319` has 27 `assign idma_req...` lines and none touch `opt.compute`; same for every variant. `src/midend/idma_nd_midend_synth.sv:347-368` likewise has no compute port.

Two consequences.

**1. Synthesis numbers for compute are meaningless.** Even elaborated with `EnableCompute=1`, `opt.compute` has no driver and constant-propagates to 0: `idma_otf_compute.sv:46` `eff_compute` becomes 0, `:51` `op_legal` becomes 0, `:53-56` all `sel_*` become 0, the passthrough mux wins, and the mxquant/mxdequant/transpose outputs go unused. The whole compute datapath is optimised away. (Chain traced in RTL; I did not measure DC behaviour on an undriven net.)

**2. The undriven field reaches live control logic.** `src/backend/tpl/idma_legalizer.sv.tpl:507-508`, generated at `target/rtl/idma_legalizer_rw_axi.sv:313-314`:

```systemverilog
decouple_rw:    req_i.opt.beo.decouple_rw | req_i.opt.compute.enable,
decouple_aw:    req_i.opt.beo.decouple_aw | req_i.opt.compute.enable,
```

These are not gated by `EnableCompute`, unlike the size-changing length calc at `:284` which is. So the undriven `opt.compute.enable` reaches the decouple flags unconditionally and any simulation of a synth wrapper yields X on `decouple_rw` / `decouple_aw`. Latent today because CI only elaborates the wrappers, it never simulates them.

Suggested fix:
- add `req_compute_*_i` ports and the matching assigns under the existing `% if compute_eligible:` guard in `idma_backend_synth.sv.tpl`
- gate `idma_legalizer.sv.tpl:507-508` with `EnableCompute &&`

Found on `devel` at `77b16cf0`; the generated evidence is from tag `v0.7.0`.

PR #234, opened 2026-09-21
frontend: Add the DMOPC compute path to the inst64 frontend

DanielKellerM:
iDMA has had in-datapath compute since #170, but `inst64` could never reach it: `idma_inst64_top` hardcoded `.EnableCompute(1'b0)` and nothing in the decode wrote `opt.compute`. So a Snitch cluster can move data but not quantise it.

`DMOPC` carries the compute opcode in `rs1` and latches it for the next transfer, so the op is selected without widening `dmcpy`.

```systemverilog
`IDMA_INST64_DMOPC(SNRT_DMA_OPCODE_MX_QUANT)   // then dmsrc / dmdst / dmcpy as usual
```

`EnableCompute`, `ComputeOps` and `ComputeTuning` become real parameters, and `CombinedShifter` follows `!EnableCompute` because the backend asserts the two are mutually exclusive.

Compute is threaded into **both** backend topologies from #233, not tied off in the OBI one. Tying it off would silently drop `opt.compute` and emit a wrong-but-successful plain copy; threaded, the legalizer decides per transfer. Both paths were checked: with AXI endpoints the OBI topology quantises byte-exact, and with a TCDM destination `ComputeMxDstProtocol` fires as it should.

`idma_inst64_compute_pkg.sv` declares only what `src/frontend/reg/idma_reg.rdl:156` encodes: passthrough `0x08`, mxquant `0x20`, mxdequant `0x21`, mxquant_fp16 `0x22`, mxdequant_fp16 `0x23`, transpose `0x50`. An undeclared byte hits `DmopcUnknownOpcode` rather than decoding to an op the hardware does not have.

## Verification

`tb_idma_inst64_compute` quantises 1024 B to 264 B and compares byte-exact against a golden, then copies back through a passthrough DMOPC. Three independent checks fail if compute is bypassed: the golden compare, a sentinel past 264 B that a plain 1024 B copy would overwrite, and a differs-from-source count.

```
-gEnableCompute=0       ComputeOpUnsupported fires; 264 mismatches;
                        "quantized block is byte-identical to the source: compute did not run"
corrupted op field      ComputeSizeAligned + ComputeMxdequantBeatAligned fire; 264 mismatches
reverted                TEST PASSED: 264 B quantized byte-exact (260 differ from the source)
```

`DmopcUnknownOpcode` has its own firing leg. `idma_lint_inst64` covers four testbenches and five `-G` sweeps including `-GEnableCompute=1`; `idma_verify_tb_shared`, `idma_verify_codegen`, `idma_lint_sv` and `lint-commits` are clean; all five inst64 sims pass on Questa 2026.1.

## Limits, documented in `snitch.md`

- Software cannot emit `DMOPC` until the opcode lands upstream; pulp-platform/riscv-opcodes#48 adds it at funct7 10. The testbench drives the accelerator bus directly, so the hardware path is exercised without it.
- Size-changing compute (MX) is AXI-only, per the legalizer's `ComputeMxSrcProtocol` / `ComputeMxDstProtocol` and the `// TODO: OBI` beside them.
- Transpose is decoded but capped at one beat by `ComputeTransposeSingleBeat`, and `idma_obi_write.sv` has no `mask_ext`, so it is not exercised here.
- No synthesis numbers: `CombinedShifter = 0` makes the write-side rotates live, and #232 must land before any compute area figure means anything.

Enabling this from a cluster needs a `cfg/*.json` knob threading `EnableCompute` down to the instantiation; that follows separately.


src/frontend/inst64/idma_inst64_snitch_pkg.sv
```diff
@@ -19,5 +19,7 @@ package idma_inst64_snitch_pkg;
   localparam logic [31:0] DMREP              = 32'b000011100000?????000000000101011;
   localparam logic [31:0] DMUSER             = 32'b0001000??????????000000000101011;
   localparam logic [31:0] DMINIT             = 32'b0001001??????????000?????0101011;
+  // Not yet allocated upstream; funct7 0x0a is the first free slot after DMINIT.
```
DanielKellerM: remove comment

src/frontend/inst64/idma_inst64_top.sv
```diff
@@ -697,6 +703,8 @@ module idma_inst64_top #(
                     idma_fe_req_d.burst_req.opt.axi_id       = idma_fe_sel_chan;
                     idma_fe_req_d.burst_req.length           = acc_req_i.data_arga;
                     idma_fe_req_d.burst_req.opt.src_protocol = idma_pkg::INIT;
+                    // the INIT read port has no compute datapath; a memset stays a memset
```
DanielKellerM: remove comment

DanielKellerM (commented):
fix

PR #161, opened 2026-07-27
Transpose/Test: Add support for compact transposition output matrices and a working verilator flow

da-gazzi:
# Compact Transposition and Verilator Testing

## Summary

- Add optional compact storage for transposed output matrices while retaining
  the existing tile-padded layout.
- Make partial transpose tiles memory-safe without changing the number or
  ordering of generated tile requests.
- Add a reusable, parameterized Verilator flow and broaden the directed
  transpose and midend regression coverage.

## Compact Transposition

- Add a software-configurable `compact` transpose option to the request and
  register interfaces.
- In compact mode, generate the destination row stride from the actual
  transposed row width. Legacy mode continues to use the tile-padded width.
- Add a request replay stage after ND expansion. Requests for inactive rows of
  partial edge tiles are redirected to the first valid source or destination
  address, preventing out-of-bounds reads and unsafe zero-strobe write
  addresses while preserving the descriptor sequence expected by the
  transpose engine.
- Clarify the transpose element-width and tile-geometry semantics in the RTL
  documentation.

## Testbench Changes

- Cover compact and padded layouts, partial edge tiles, aligned and unaligned
  matrices, 1-, 2-, and 4-byte elements, multiple bus widths, backpressure,
  and back-to-back transfers.
- Check results against the DPI golden model and verify sentinel regions and
  generated write-address bounds.
- Exercise full- and half-duplex transpose configurations together with the
  ND, runtime, and register frontend midends.
- Use SystemVerilog clocking blocks for race-free ready/valid stimulus and
  handshake observation.

## Verilator Flow

- Add common Verilator build and run plumbing with independent build
  directories for each testbench parametrization.
- Generate stable Bender file lists and track RTL sources, headers, DPI code,
  and elaboration options so binaries rebuild when their inputs change.
- Provide explicit per-testbench parameter matrices, convenient aggregate
  targets, clean targets, optional FST tracing, and configurable compiler and
  linker flags.
- Make the transpose DPI model link correctly when compiled as C++ by
  Verilator.

## Validation

- The directed compact/padded transpose configurations, back-to-back tests,
  on-the-fly transpose variants, and affected midend tests pass with the new
  Verilator flow.

# TODO
- The non-free gitlab CI does not pass - as I am not at IIS anymore, I can't see why, I will happily change whatever is needed to fix this.

PR #239, opened 2026-09-22
backend: Advance the transpose engine on the write-beat handshake

DanielKellerM:
The transpose sub-unit took its output ready from `w_dp_req_ready`, while the MX units correctly use the per-beat write pop. On AXI and TileLink that signal is burst-granular (`idma_axi_write.sv:221`, `w_dp_ready_o = last_w & write_happening`), so a multi-beat burst produced one ready pulse for N beats: `rd_cnt` never advanced, every output beat replayed column 0, the bank never reached `drain_done`, and the read side then wedged. OBI, INIT and AXI-Stream drive `w_dp_ready_o = write_happening` and were accidentally correct, which is why only AXI destinations broke.

`ready_i` on `idma_otf_transpose` is an ordinary beat-granular stream ready: it feeds `ready_int` and, through `out_hs`, drives `rd_cnt`, which selects the transposed read-out column and the edge strobe. One pulse must mean one output beat consumed.

Drive it from `w_chan_valid_o & w_chan_ready_o` instead. Not `buffer_out_ready_shifted`: on the last column-tile the transpose emits all-zero `strb_o` for padding rows of the transposed matrix, so `mask_out` and hence `buffer_out_ready` are zero even though the bus beat was accepted, and reducing that signal would hang on every masked edge beat. All five write ports export `w_chan_valid_o`/`w_chan_ready_o` identically and the transport layer already muxes them.

`ComputeTransposeSingleBeat` becomes `ComputeTransposeShape`. The single-beat restriction was containing this bug, but order still constrains the shape: the engine consumes (col-tile, row-tile, row) and emits transposed rows at pitch `MP*E`, which a contiguous burst can only express for a single tile. The relaxation is exactly `length == NE*StrbWidth && M <= NE && N <= NE`; everything else stays rejected. `tb_idma_mxneg` case 10 still fires.

This also unblocks the frontends that have no transpose midend. `idma_inst64_top` fixes `NumDim = 2`, so software had to issue one request per `StrbWidth` strip; with `dma_data_width: 512` an int8 tile needs 64 input beats before the engine emits anything while the request queue holds ~8, which deadlocks on capacity. One request per tile removes that. The register frontend is affected identically and gets the same fix.

Regression test: six whole-tile single-burst cases in `tb_idma_transpose_nd`. All non-square, so an M/N swap is visible; bytes are position-coded and the test additionally requires exactly `(N*M - min(M,N))*EB` bytes to differ from what a plain copy would leave, so a passthrough scores 0 and fails. Before the fix three cases mismatch with the column-0-replicated signature and `2x3 EB=2` trips the watchdog; after, all pass.

Verified on Questa 2026.1 with work libs wiped: `tb_idma_transpose_nd` 13 walk + 6 tile cases at StrbWidth 4 and 8, `tb_idma_otf_transpose` against the DPI-C golden across 4 cases x 2 widths x 2 duplex settings, `tb_idma_transpose_midend`, `tb_idma_transpose_b2b`, `tb_idma_inst64_compute`, the full MX set (`mxquant`, `mxroundtrip`, `mxrand`, `mxneg`, `mxclear`), and backend `rw_axi`/`r_obi_w_axi`. `make idma_verify_codegen`, verible and `idma_lint_inst64` clean.


DanielKellerM:
Closing in favour of #240, which carries Georg's lane-consumption accumulator from #161.

This fix is not merely superseded, it is wrong. Driving the transpose retire from `w_chan_valid_o & w_chan_ready_o` fixes the burst-granular under-advance but reopens the misaligned-destination case Georg fixed in `23b95c0c`: a misaligned row spans two bus beats, the pulse fires twice, and `rd_cnt` over-advances. Measured against a misaligned sweep it does not merely corrupt, it deadlocks at both bus widths on the first misaligned case, a shape that works on devel today:

```
[TPN] PASS: 13x19 EB=1 misalign=0
[TPN] case 8x8 EB=1 (NE=4, 2x2 tiles)
** Fatal: Tripped Watchdog ... Inactivity for 2000 cycles
```

Nothing guarded it: `ComputeDstAligned` is gated on `compute_in_bytes != compute_out_bytes`, and transpose keeps its size, so it is exempt. Nothing tested it either, since both transpose testbenches used an aligned destination base.

The accumulator covers all four cases instead: aligned multi-beat, misaligned two-beat rows, 4 KiB splits, and zero-strobe edge beats.

The additive half of this PR survives and moves to a follow-up on top of #240: the `ComputeTransposeSingleBeat` to `ComputeTransposeShape` relaxation that lets one request carry a whole tile, and the whole-tile regression cases. That relaxation is what unblocks the Snitch side, where `idma_inst64_top` fixes `NumDim = 2` with no transpose midend, so software had to issue one request per strip and deadlocked on request-queue capacity.

If #240 stalls, the smaller fallback is to lift `04676c24` alone rather than revive this branch.

PR #241, opened 2026-09-22
backend: Allow a whole transpose tile in one multi-beat burst

DanielKellerM:
Follow-up to #240, which landed Georg's lane-consumption accumulator. This is the additive half of the now-closed #239, rebuilt on that handshake instead of its own.

**Why the relaxation matters.** `ComputeTransposeSingleBeat` rejected every transpose request longer than one beat. A frontend without a transpose midend has to tile in software, and `idma_inst64_top` fixes `NumDim = 2`, so Snitch had to issue one request per `StrbWidth` strip. At `dma_data_width: 512` an int8 tile needs 64 input beats before the engine emits anything while the request queue holds about 8, so it deadlocked on capacity. One request per tile removes that.

`ComputeTransposeShape` additionally accepts one whole padded tile (`length == NE * StrbWidth`, `M` and `N` both `<= NE`). Everything else stays rejected. No handshake change is needed: #240's accumulator already retires an output beat only once every logical byte lane has been consumed.

**A multi-beat burst does require a beat-aligned destination, and this is a real limit, not caution.** The write barrel shifter rotates and the compute output is a single registered beat, so a misaligned bus beat would have to draw from two consecutive compute beats at once. The intermediate beats of such a burst carry an unmasked logical mask, claim all `StrbWidth` lanes after rotation, and retire the engine a beat early, shifting the whole tile by one destination row. A 4-way isolation sweep separated the axes cleanly: all 12 source-misaligned cases pass, all 12 destination-misaligned fail. `ComputeDstAligned` does not cover it because it is gated on `compute_in_bytes != compute_out_bytes` and transpose keeps its size. The shape guard now requires it, with `mxneg` case 14 as the negative test. Single-beat strips stay unrestricted at any alignment, and a misaligned source stays legal in both shapes since the per-lane dataflow buffer absorbs it.

Note Georg's two SVAs (`ComputeConsumeValid`, `ComputeBeatAllLanesValid`) survive the relaxed shape and do not fire, but they also cannot catch this: the lanes genuinely are all covered and all valid, what is wrong is which compute beat they belong to. Hence the legalizer.

**Second commit** fixes a silent test gap found on the way. All three transpose testbenches derived the element-size mode with a ternary missing `eb == 8`, so EB=8 was programmed as mode 0 and never exercised. Fixed in all three, with EB=8 geometries added and an aligned/sub-beat-misaligned base sweep in the nd testbench. That sweep is what demonstrated the destination limit above.

Verified on Questa 2026.1 with work libs wiped: `tb_idma_transpose_nd` (`ALL PASS: 14 walk x 2 alignments + 6 tile cases x 3 variants` - aligned, misaligned source, and a beat-aligned destination that splits on a 4 KiB boundary), `tb_idma_transpose_b2b`, `tb_idma_mxneg` (cases 10 and 14 both FIRED), `tb_idma_inst64_compute`. Plus `make idma_verify_codegen`, verible and `idma_lint_inst64`, all clean, and the working tree stays clean across a full build.


doc/site/src/content/docs/architecture/compute.md
```diff
@@ -57,7 +57,7 @@ The register frontend exposes these through its `compute_cfg` register. The op e
...
+Transpose does not change transfer size (input bytes == output bytes). The transport layer retires an output beat only once every logical byte lane of that beat has been consumed by the write port (`cmp_consumed_q`), so one output beat may span several bus beats (a misaligned destination, a 4 KiB split) without over-advancing the engine. The legalizer (`ComputeTransposeShape`) accepts two shapes: a tiled-walk strip of at most one beat (`length <= StrbWidth`, the shape the midend emits), or one whole padded tile in a single burst (`length == NE * StrbWidth` with `M <= NE` and `N <= NE`), where the source is stored at a `StrbWidth` row pitch. The multi-beat shape additionally requires a beat-aligned destination: the write barrel shifter rotates, so a misaligned bus beat would draw from two consecutive compute beats at once, which the single registered compute output cannot serve. The source may be misaligned in either shape, since the per-lane dataflow buffer absorbs that. Any other multi-beat shape would need the tiled (col-tile, row-tile, row) feed order that a contiguous burst cannot express.
```
DanielKellerM: this is quite verbose, must be 3 max consicse sentences

src/backend/tpl/idma_legalizer.sv.tpl
```diff
@@ -736,10 +746,17 @@ ${database[protocol]['legalizer_write_data_path']}
     // compute retires on the per-beat write pulse; TileLink writes retire per burst
     `ASSERT_NEVER(ComputeDstTilelink, (ready_o & valid_i & req_i.opt.compute.enable &
                   (req_i.opt.dst_protocol == idma_pkg::TILELINK)), clk_i, !rst_ni)
-    // NOT IMPLEMENTED: multi-beat transpose write bursts (midend strips are single-beat)
-    `ASSERT_NEVER(ComputeTransposeSingleBeat, (ready_o & valid_i & req_i.opt.compute.enable &
+    // transpose shapes: a tiled-walk strip (<= one beat) or one whole padded NE x NE tile.
+    // A multi-beat burst needs a beat-aligned destination: the write shifter rotates, so a
+    // misaligned bus beat would straddle two compute beats, which the datapath cannot serve.
```
DanielKellerM: verbose block, must be max 1 line

test/tb_idma_mxneg.sv
```diff
@@ -104,6 +104,9 @@ module tb_idma_mxneg
                 idma_pkg::AXI, idma_pkg::AXI, 1'b0);
       13: issue(Src, Dst, 64, idma_pkg::COMPUTE_MXQUANT_FP16,
                 idma_pkg::AXI, idma_pkg::AXI, 1'b0);
+      // whole-tile length, but a misaligned destination the write shifter cannot serve
```
DanielKellerM: remove comment

